PR #1, opened 2015-02-28
Fail on requests with unknown headers matching X-Thentos-*.

mf59816:
(See comment in the test case on why this is a 500 error and not something more sane.)

### DIFF
--- a/src/Thentos/Backend/Core.hs
+++ b/src/Thentos/Backend/Core.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE DataKinds                                #-}
+{-# LANGUAGE DeriveDataTypeable                       #-}
 {-# LANGUAGE ExistentialQuantification                #-}
 {-# LANGUAGE FlexibleContexts                         #-}
 {-# LANGUAGE FlexibleInstances                        #-}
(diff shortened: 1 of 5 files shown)


src/Thentos/Backend/Core.hs
```diff
@@ -93,3 +96,53 @@ fmapLTM trans e = EitherT $ do
...
+
+renderThentosHeaderName :: ThentosHeaderName -> CI SBS
+renderThentosHeaderName x = case splitAt (SBS.length "ThentosHeader") (show x) of
+    (_, s) -> mk . SBS.pack $ "X-Thentos" ++ dashify s
+
+    -- (oddness: it would be strictly better to use the pattern
+    -- @("ThentosHeader", s)@ here: if "ThentosHeader" does not match,
+    -- we *want* this to crash, but we are pretty confident that it
+    -- won't.  but if we do not use an @_@ here, we get an
+    -- unmatched-pattern warning.)
```
fhartwig: Why not have an explicit "This should not happen" branch?

mf59816: Because it's UGLY!  :-)

Ok, I'll do that anyway.

PR #2, opened 2015-03-04
Add login/logout benchmark

fhartwig:
This benchmark has the same problem as the signup one, namely that it spends almost all of its time in scrypt.

### DIFF
--- a/bench/Main.hs
+++ b/bench/Main.hs
@@ -4,7 +4,7 @@ module Main (main) where
 
 import Data.Aeson (encode, decode)
 import Data.List (unfoldr)
-import Data.Maybe (fromJust)
+import Data.Maybe (fromJust, fromMaybe)
 import Data.Monoid ((<>))
 import Data.Text.Encoding (encodeUtf8)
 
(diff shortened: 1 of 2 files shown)


mf59816:
Do you also get figures like this?:

    Benchmark load-test: RUNNING...
    foo in 49.31 sec (3.6% CPU)
    49.312523
    latency:
        mean:    4.585 sec
        std dev: 1.024 sec
        99%:     6.116 sec
        99.9%:   6.592 sec

    throughput:  20.28 req/sec
    foo in 25.88 sec (6.2% CPU)
    25.877134
    latency:
        mean:    2.442 sec
        std dev: 1.938 sec
        99%:     7.758 sec
        99.9%:   7.889 sec

    throughput:  38.64 req/sec
    Benchmark load-test: FINISH

(Running thentos in the interpreter, single-threaded.)

Why is logging in slower than logging in and creating a user?

fhartwig:
I get similar numbers.
In the first benchmark, every request creates a user (so there is one call to scrypt per request). The second benchmark consists of alternating login and logout requests, so only half of the requests need to call scrypt.

mf59816:
Done with code review.  I like this branch.

Please update what you think should be updated and re-assign to me.

PR #5, opened 2015-03-05
Handle attempts to create new user with existing email address

fhartwig:
When someone tries to sign up with an email address that we already have in the database, we now send a "By the way, you're already signed up for Thentos" email instead of creating an unconfirmed user with the same email address. However, the user still gets the same response from the frontend, to avoid leaking user data.


mf59816:
I would like to keep to the branch names with date and user prefix, and push everything to liqd/thentos.  Rationale:

1. I like the branch names.  They are *organised*!  (-:
2. Anybody can push updates on anybody else's pull request if that's easier than writing a coment and asking the branch author to do it.

(@fhartwig: you already have write access to liqd/thentos, in case you are wondering.)

mf59816:
How much effort would it be to write tests for this?

mf59816:
Test works.  Please process comments and assign back to me.  Thanks!

PR #7, opened 2015-03-06
Add first (and quite arbitrary) set of rules.

mf59816:
Please object where necessary!

### DIFF
--- a/docs/styleguide.md
+++ b/docs/styleguide.md
@@ -5,9 +5,23 @@ List of programming style rules
 (will be re-structured as it grows)
 
 - qualified imports are given the name of the module (without path),
-  not an abbreviation (`import qualified Data.Map as Map`).
+  not an abbreviation.  good:
 
-- exceptions:
+    import qualified Data.Map as Map
+
+  bad:
+
+    import qualified Data.Map as M
+
+  obviously, this rule sometimes needs bending.  bad:
+
+    import qualified Data.HashMap.Strict as Strict
+
+  good:
+
+    import qualified Data.HashMap.Strict as HashMap
+
+  further exceptions:
 
     import Data.ByteString as SBS
     import Data.ByteString.Lazy as LBS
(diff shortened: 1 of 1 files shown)


docs/styleguide.md
```diff
@@ -0,0 +1,24 @@
+
+List of programming style rules
+===============================
+
+(will be re-structured as it grows)
+
+- qualified imports are given the name of the module (without path),
+  not an abbreviation (`import qualified Data.Map as Map`).
+
```
fhartwig: This is going to lead to hugely unhelpful import names in many cases (e.g. `Data.Map.Lazy` and friends).

docs/styleguide.md
```diff
@@ -0,0 +1,24 @@
...
+- exceptions:
+
+    import Data.ByteString as SBS
+    import Data.ByteString.Lazy as LBS
+    import Data.Text as ST
+    import Data.Text.Lazy as LT
+
+- type synonyms SBS, LBS, ST, LST from string-conversions are used in
+  type signatures (even if no strings are converted).
+
```
fhartwig: I don't love this rule. I don't think it improves anything, but it makes signatures less readable by replacing well-known types by obscure and unreadable synonyms.

docs/styleguide.md
```diff
@@ -0,0 +1,24 @@
...
+- maximum line length is 140 chars.
```
fhartwig: 140 chars is huge. At that point, why have a rule at all?

docs/styleguide.md
```diff
@@ -0,0 +1,24 @@
...
+
+- preceed haddock section headings with two empty lines.
+
+- either values are named `eSession`; maybe values `mSession`.
```
fhartwig: This might be confusing. Also, what if I have two `Either`s in the same scope? ;-)

docs/styleguide.md
fhartwig: This would also prevent us from doing things like

    import qualified Data.Aeson as Aeson
    import qualified Data.Aeson.Encode.Pretty as Aeson
    import qualified Data.Aeson.Types as Aeson

(your code, not mine)

mf59816: fair enough (both comments).  what i want to avoid is single-character imports like `B`, `C`, `A`.  i'll think of a rule that makes more sense.

docs/styleguide.md
mf59816: 140 /= NaN!  (-:

docs/styleguide.md
```diff
@@ -0,0 +1,68 @@
...
+
+## naming
+
+Names for `Either` values start with an `e` (example: `eSession`).
+Analogously, `Maybe` value names start with an `m`.
+
+
+## string types
+
+type synonyms SBS, LBS, ST, LST from string-conversions are used in
```
fhartwig: `LST` should be `LT`, I guess?

PR #8, opened 2015-03-10
Add password reset feature to frontend

fhartwig:


mf59816:
i'm going to review (and hopefully merge) this without bothering about the travis issue.

mf59816:
Ok, this is good, BUT: many pedandic details (see above), and two general points:

1. there are no tests.  we need frontend-tests.  they don't have to be webdriver-based; we can construct the requests that the browser would send manually and observe the responses manually.  on the other hand, it may even be easier to start with webdriver right away.  it's ugly and buggy in the corner cases, but it's relatively high-level.

2. the authorization system is not really used: all labels are `thentosPublic`, and all clearances that need to be stated explicitely are `allowEverything`.  this needs some more work, preferably in this PR.

(1) can be moved to a separate pull request.  i'm ok with taking over any one or both of these points if you want me to.  there is enough other stuff for you to do.

but could you go through the list of comments and process (resp. argue against) them?

fhartwig:
re: 2:
That's why I asked you to look at the permissions stuff in particular, but I was hoping for something more specific than "this needs work".
re: 1:
You're right, there should be tests. I've never seen webdriver tests that were even close to being useful enough to justify the effort invested in them, so I'd rather avoid them as long as we can.

mf59816:
I rewrote this to what I think is a bit clearer.  what do you think?

Also, please merge with current master and resolve conflicts.

PR #10, opened 2015-03-17
Support for webdriver / selenium.

mf59816:
(not ready for merge yet.  still testing / debugging.)

### DIFF
--- a/misc/selenium/Makefile
+++ b/misc/selenium/Makefile
@@ -28,7 +28,7 @@ SELENIUM_HUB_PORT=4451
 SELENIUM_NODE_PORT=4452
 # (default: 5555)
 
-SELENIUM_ARGS=-jar selenium-server-standalone-2.45.0.jar -host $(SELENIUM_HOST) -port $(SELENIUM_HUB_PORT)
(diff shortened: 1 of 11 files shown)


mf59816:
right: travis.  oh dear.

mf59816:
i'll look into how we can get this to work with travis.  @fhartwig: would you review / reject / merge this independently of travis, and we'll fix this elsewhere?

misc/selenium/Makefile
```diff
@@ -0,0 +1,55 @@
...
+default:
+	cat Makefile
+	make kill start watch
+
+init:
+	curl http://selenium-release.storage.googleapis.com/2.45/selenium-server-standalone-2.45.0.jar > selenium-server-standalone-2.45.0.jar
+	mkdir -p log
+
+kill:
+	killall java > /dev/null 2>&1 || true
```
pallix: what about users who run other java processes :-) ?

mf59816: they run java.  their lives can only improve after this.  ;-)

yes, i know, this is all not very sturdy.  would you like to patch it?

(@fhartwig also wants to add a flag to disable selenium tests; then at least it will be easy (or even the default) to get away from this mess entirely.)

PR #15, opened 2015-03-18
Fix travis

mf59816:
(this is an answer to 
https://github.com/liqd/thentos/commit/bbf2d593db38d17533a2eca9b18c9d0c5138c3c7#commitcomment-10270895)


mf59816:
I've added Makefile rules for testing, where the simpler rule is the one without selenium.  The problem with making "without selenium" the default is that it's not clear to me how to do that with hspec (which was your essentially your point in our discussion earlier today).  On the other hand, I think default should be that all tests are run, and if you want to skip tests (even for good reasons) you should have to say so.

Does that by any chance convince you...?

PR #17, opened 2015-03-24
Configifier

mf59816:
There is still room for improvement, but this is my fist shot.

Next steps (on other branches):

- use more specific types in config structure.
- turn lookup errors under helpers into parse errors.


mf59816:
oops, wrong branch.  sorry.

PR #18, opened 2015-03-24
Move from configurator, optparse-applicative to configifier.

mf59816:
There is still room for improvement, but this is my fist shot.

### DIFF
--- a/devel.config
+++ b/devel.config
@@ -2,11 +2,11 @@ command: "run"
 
 backend:
     bind_port: 7001
-    bind_host: "localhost"
+    bind_host: "127.0.0.1"
 
 frontend:
     bind_port: 7002
-    bind_host: "localhost"
+    bind_host: "127.0.0.1"
 
 smtp:
     sender_name: "Thentos"
--- a/devel.config
+++ b/devel.config
@@ -1,16 +1,21 @@
-backend_port = 7001
-frontend_port = 7002
+command: "run"
 
-smtp {
-    sendmail_path = "/usr/sbin/sendmail"
-    sendmail_args = ["-t"]
-    sender_name = "Thentos"
-    sender_address = "thentos@thentos.org"
-}
+backend:
+    bind_port: 7001
+    bind_host: "localhost"
 
-default_user {
-    name = "god"
-    password = "god"
-    email = "postmaster@localhost"
-    roles = ["RoleAdmin"]
-}
+frontend:
+    bind_port: 7002
+    bind_host: "localhost"
+
+smtp:
+    sender_name: "Thentos"
+    sender_address: "thentos@thentos.org"
+    sendmail_path: "/usr/sbin/sendmail"    # (built-in default)
+    sendmail_args: ["-t"]                  # (built-in default)
+
+default_user:
+    name: "god"
+    password: "god"
+    email: "postmaster@localhost"
+    roles: ["RoleAdmin"]
--- a/devel.config
+++ b/devel.config
@@ -18,4 +18,4 @@ default_user:
     name: "god"
     password: "god"
     email: "postmaster@localhost"
-    roles: ["RoleAdmin"]
+    roles: ["roleAdmin"]
--- a/src/Thentos/Config.hs
+++ b/src/Thentos/Config.hs
@@ -1,311 +1,182 @@
-{-# LANGUAGE OverloadedStrings                        #-}
-{-# LANGUAGE ScopedTypeVariables                      #-}
-{-# LANGUAGE TupleSections                            #-}
+{-# LANGUAGE DataKinds            #-}
+{-# LANGUAGE OverloadedStrings    #-}
+{-# LANGUAGE ScopedTypeVariables  #-}
+{-# LANGUAGE TupleSections        #-}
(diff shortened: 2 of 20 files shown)


mf59816:
The following have been addressed now (removing them from the original pr comment):

- use more specific types in config structure.
- turn lookup errors under helpers into parse errors.

devel.config
```diff
@@ -1,16 +1,21 @@
-backend_port = 7001
-frontend_port = 7002
+command: "run"
```
fhartwig: Would it be possible to always take the command from the command line? Taking it commands from a config file is a bit weird, I think.
If not, no big deal.

src/Thentos/Config.hs
```diff
@@ -1,312 +1,199 @@
...
+    sources <- sequence
+        [ ConfigFileYaml <$> SBS.readFile configFile
+        , ShellEnv       <$> getEnvironment
+        , CommandLine    <$> getArgs
+        ]
+    logger DEBUG $ "config sources:\n" ++ ppShow sources
+
+    case configify' (TaggedM defaultThentosConfig) sources :: Result ThentosConfigUntagged of
+        Left e -> do
+            logger CRITICAL $ "error parsing config: " ++ show e
```
fhartwig: I currently get this error when doing a `cabal run -- thentos` (with the devel.config unchanged). Is there something I've missed?

src/Thentos/Util.hs
```diff
@@ -77,3 +83,13 @@ readsPrecEnumBoundedShow _ s = f [minBound..]
         (s0, s1) -> if s0 == s' then [(x, s1)] else f xs
       where
         s' = show x
+
+
```
fhartwig: This seems to be only used in tests. Maybe it should be moved to the tests Util module?

fhartwig:
Looks good. Some comments, but the only major thing is that I can't get Thentos to run due to some config parse error.

devel.config
mf59816: you can choose not to set the command in the config file, or you can override a default command on the command line, like this:

    thentos --command=runA3

proper sub-commands (as git knows them up to 3 levels or so deep) will be implemented at some point.

(it may also be a reasonable requirement in some settings to "ban" specific sub-configs from specific sources as you suggest.  i will put that on my list, but unless i have a good use case it will probably not make it to the top.)

fhartwig: Ok, cool, that's good enough for now. I almost always run `run` anyway. It's not like Thentos is git or cabal.

src/Thentos/Config.hs
mf59816: having the error message would help.  (-:

fhartwig:
```
error parsing config: InvalidYaml {invalidYamlInput = "command: \"run\"\n\nbackend:\n    bind_port: 7001\n    bind_host: \"127.0.0.1\"\n\nfrontend:\n    bind_port: 7002\n    bind_host: \"127.0.0.1\"\n\nsmtp:\n    sender_name: \"Thentos\"\n    sender_address: \"thentos@thentos.org\"\n    sendmail_path: \"/usr/sbin/sendmail\"    # (built-in default)\n    sendmail_args: [\"-t\"]                  # (built-in default)\n\ndefault_user:\n    name: \"god\"\n    password: \"god\"\n    email: \"postmaster@localhost\"\n    roles: [\"RoleAdmin\"]\n", invalidYamlMsg = "Error parsing enumeration: expected roleOwnsUnconfirmedUsers, found RoleAdmin."}
```
I assumed you'd get the same error on this branch.

src/Thentos/Util.hs
mf59816: what about this?

    $ git grep readsPrecEnumBoundedShow
    src/Thentos/Backend/Api/Adhocracy3.hs:    readsPrec = readsPrecEnumBoundedShow
    src/Thentos/Backend/Api/Adhocracy3.hs:    readsPrec = readsPrecEnumBoundedShow

fhartwig: Sorry, I was referring to `<//>`.

fhartwig: Also, "concatenization" is not a word :-p

src/Thentos/Config.hs
mf59816: Yes, I am now, but you just had it right in front of you, and it took me 5 minutes to check out the branch build it again, and get the error myself.  I don't know the answer to this error immediately, so I had to go through that anyway, but you didn't know that.

Anyway, my bad.  I submitted a branch for review appearently without having started the main executable, and with a sketchy test suite....

I'll look into it.

src/Thentos/Util.hs
mf59816: (-:

I would prefer to keep `<//>` in the lib.  It should always be used instead of `<>` used for concatenizationalizing paths, so we don't have to check manually for leading or trailing slashes.

mf59816:
erm, that wasn't too hard, could have done it right away.  (configifier error messages could use some work, but that's at least partly due to the underlying attoparsec and yaml package.  it's on my configifier TODO list now.)

PR #21, opened 2015-03-26
Cleanup.

mf59816:
This should be easy to review.  (-:


mf59816:
This looks like a travis problem.  Could you test locally and decide whether to merge based on that?

PR #22, opened 2015-03-28
Cabal hacks

mf59816:
Misc cleanup changes in `thentos.cabal`, `cabal.config`.


mf59816:
Travis now takes 4 minutes with up-to-date cache blob.  I'm a bit disappointed, but it's better than before.

fhartwig:
Looks good. This means that we manually have to update the cache when our deps change, right (as opposed to travis updating the cache for us)? We might at some point want to automate that, but this is definitely a huge improvement.

PR #23, opened 2015-03-30
Cleanup Frontend

fhartwig:


mf59816:
I added the rest of the changes we discussed.  Now there is an error in the wd tests (which luckily won't show on travis :-).

I also managed to reproduce it manually on the running system, but got very little feedback on where the error could lie.  @fhartwig, could you take a look?  My guess is that some of my renamings are inconsistent.  thx-

mf59816:
I tried some more: Creating users in the frontend manually works.  The wd test fails on submitting the user form with this error from acid-state:

    Access failure: Core closed.

This is really weird since this branch didn't really touch anything related to acid-state.

I suggest we merge this and, if the problem persists, fix it in a separate branch.

mf59816:
(Btw, do you have this problem now, or did you before?  I don't think it has to do with my commit any more.)

fhartwig:
I haven't seen this problem, but that may be because I don't usually run the webdriver tests. I'll check if I can reproduce it on my desktop and have a go at debugging it. I don't love the idea of merging something with known-to-be-broken tests.

fhartwig:
The webdriver tests now pass for me. There were just some form field ids that needed to be fixed after we moved from `/create_user` to `user/create". I can't reproduce any acid-state-related issues.

PR #24, opened 2015-03-31
Frontend Tests

mf59816:
As discussed.

There is a one-line merge conflict with #23 where I had to change a name in `FrontendSpec`.  Should be easy to resolve.  You can process #23 first, then merge master into this.


fhartwig:
There is some minor work left to do on the service creation functionality and logout is still not implemented, but this is in a state that can be merged, I think.

PR #25, opened 2015-04-10
2015 04 08 fh update user

fhartwig:
Don't merge this yet, this branch still needs some cleaning-up and work on tests.


tests/ThentosSpec.hs
```diff
@@ -69,13 +69,17 @@ spec = do
...
     describe "UpdateUser" $ do
       it "changes user if it exists" $ \ (st, _, _) -> do
-        result <- update' st $ UpdateUser (UserId 1) user1 allowEverything
+        result <- update' st $ UpdateUserField (UserId 1)
+                                               (UpdateUserFieldName "fka_user1")
+                                               allowEverything
         result `shouldBe` Right ()
         result2 <- query' st $ LookupUser (UserId 1) allowEverything
-        result2 `shouldBe` (Right (UserId 1, user1))
+        result2 `shouldBe` (Right (UserId 1, user1 {_userName = "fka_user1"}))
```
mf59816: or, equally valid:

    result2 `shouldBe` (Right (UserId 1, userName .~ "fka_user1" $ user1))

src/Thentos/Smtp.hs
```diff
@@ -6,6 +6,7 @@ module Thentos.Smtp
     ( sendUserConfirmationMail
     , sendUserExistsMail
     , sendPasswordResetMail
+    , sendEmailChangeConfirmMail
```
mf59816: `sendChangeConfirmMail` would be more consistent with the existing names.

src/Thentos/Frontend/Handlers.hs
```diff
@@ -110,20 +111,90 @@ userCreateConfirm = do
...
+                Left NoSuchToken -> blaze $ errorPage "No such token"
+                Left e -> do
+                    logger WARNING (show e)
+                    blaze $ errorPage "Error when trying to change email address"
+
+runWithUserClearance :: (ThentosClearance -> UserId -> FH ()) -> FH ()
 runWithUserClearance = runWithUserClearance' ()
 
-runWithUserClearance' :: a -> (ThentosClearance -> FH a) -> FH a
+runWithUserClearance' :: a -> (ThentosClearance -> UserId -> FH a) -> FH a
```
mf59816: the fact that the function argument also wants a uid should probably be discussed in a comment.  (answer questions like: how is it filled?  why is this not done before the call to `runWithUserClearance`?)

src/Thentos/Frontend/Handlers.hs
```diff
@@ -110,20 +111,90 @@ userCreateConfirm = do
...
+    (emailView, result) <- runForm "update_email" emailUpdateForm
+    case result of
+        Nothing -> blaze $ userUpdatePage userView emailView passwordView
+        Just newEmail -> do
+            result' <- snapRunAction' clearance $ requestUserEmailChange uid newEmail
+            case result' of
+                Right token -> do
+                    config :: ThentosConfig <- gets (^. cfg)
+                    feConfig <- gets (^. frontendCfg)
+                    liftIO $ sendEmailChangeConfirmMail (Tagged $ config >>. (Proxy :: Proxy '["smtp"]))
```
mf59816: send out emails in `Thentos.Api`?  the benefit would be that the backend does not have to duplicate that code if email update is every triggered via rest.  (the idea is that thentos can be accessed both via the snap api and -- alternatively -- entirely via rest.)

src/Thentos/Frontend.hs
```diff
@@ -58,7 +58,10 @@ routes = [ ("", ifTop $ H.index)
          , ("user/create_confirm", H.userCreateConfirm)
          , ("user/reset_password_request", H.resetPasswordRequest)
          , ("user/reset_password", H.resetPassword)
-         -- , ("user/update", ?)
+         , ("user/update", H.userUpdate)
+         , ("user/update_email", H.emailUpdate)
+         , ("user/email_confirm", H.emailUpdateConfirm)
```
mf59816: `"user/update_email_confirm"` instead?

src/Thentos/DB/Trans.hs
```diff
@@ -244,38 +264,36 @@ trans_resetPassword now token newPass = do
...
 
+-- | Update existing user in DB.  Throw an error if user id does not
+-- exist, or if email address in updated user is already in use by
+-- another user.
 trans_updateUserField :: UserId -> UpdateUserFieldOp -> ThentosUpdate ()
-trans_updateUserField uid op = do
+trans_updateUserField uid op = trans_updateUserFields uid [op]
+
+trans_updateUserFields :: UserId -> [UpdateUserFieldOp] -> ThentosUpdate ()
+trans_updateUserFields uid ops = do
```
mf59816: nice idea.  this way we can still choose to change any number of fields inside one transaction, but we don't have to.  this makes `updateUser` completely unnecessary.

src/Thentos/Smtp.hs
fhartwig: How so? It's `send Password-reset mail`, why not `send email-change confirm mail`? (Although it should be "confirmation", not "confirm")

tests/ThentosSpec.hs
fhartwig: Yeah, that's better. I'm really bad at remembering lens operators :disappointed:

src/Thentos/Smtp.hs
mf59816: you're right.  please don't change it.  (i don't have an opinion on 'confirmation' vs. 'confirm'.)

src/Thentos/Frontend.hs
fhartwig: :+1:

src/Thentos/Frontend/Handlers.hs
fhartwig: You're right, I'll do that.

tests/ThentosSpec.hs
mf59816: you can always go to the import list in `Thentos.Trans.DB` and then search the module for how the operators mentioned there are used.  that should get you far enough.  (-:

mf59816:
replaced by #28

PR #29, opened 2015-04-13
Make email confirmations tokens expire

fhartwig:


mf59816:
(ran test suite locally.  looks good.)

PR #31, opened 2015-04-13
Dashboard pagelet

mf59816:
Comes with a very simple demonstration.  Missing: body pages for all tabs (to be done separately?).


fhartwig:
Two minor points, looks good.

mf59816:
I decided to leave both points unchanged: Typeable is implicit starting from ghc 7.10 anyway, and the `(==) ->` is an idiom I think is worth learning.  It's sooo pretty!

(Sorry for being so stubborn. :- )

PR #38, opened 2015-04-14
Refactor DB type

mf59816:
The last commit explains that the transactions still work as before.  I would like to update them separately.


fhartwig:
Looks good. :+1:

PR #42, opened 2015-04-20
Split up roles into two types (maintains representation in config files)

mf59816:
also fixes bug introduced unnoticed a bit earlier.


mf59816:
test suite still fails.  i'll look into it.

PR #47, opened 2015-04-25
Frontend Misc

mf59816:
- user registering with service
- snap error handling
- use uri-bytestring package
- various misc refactorings and bug fixes

### DIFF
--- a/src/Thentos/Frontend/Handlers.hs
+++ b/src/Thentos/Frontend/Handlers.hs
@@ -17,7 +17,7 @@ import Data.Configifier ((>>.), Tagged(Tagged))
 import Data.Functor.Infix ((<$$>))
 import Data.Monoid ((<>))
 import Data.Proxy (Proxy(Proxy))
-import Data.String.Conversions (cs)
+import Data.String.Conversions (SBS, cs)
 import Data.Text.Encoding (decodeUtf8, decodeUtf8', encodeUtf8)
 import LIO.DCLabel ((/\), (\/))
 import Snap.Blaze (blaze)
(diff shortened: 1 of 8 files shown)


mf59816:
(Travis is expected to fail.  See comment in import section in `Thentos.Frontend.Handlers`.  I'm a bit confused about the error travis actually throws, but I'll look into this in a different PR once uri-bytestring-0.1 is out.)

fhartwig:
@mf59816 The travis error is because we haven't added `uri-bytestring` to our dependency cache, right?

src/Thentos/Frontend/Handlers.hs
```diff
@@ -1,47 +1,56 @@
...
+import Snap.Snaplet.Session (commitSession, setInSession, getFromSession, resetSession)
 import Snap.Snaplet (with)
+import System.Log.Missing (logger)
 import System.Log (Priority(DEBUG, INFO, WARNING, CRITICAL))
 import Text.Digestive.Snap (runForm)
+import URI.ByteString (URI(..), RelativeRef(..), URIParserOptions)
+import URI.ByteString (parseURI, parseRelativeRef, laxURIParserOptions, serializeURI, serializeRelativeRef)
+import URI.ByteString (rrPathL, uriQueryL, queryPairsL)
+
+-- We are using an experimental version of URI.ByteString:
```
fhartwig: Could we not add this version to our travis dependency cache, since we just copy the entire sandbox anyway?

mf59816: yes, travis doesn't find this in the cache, but the script was supposed to fall back to installing missings.

I would prefer not to fix this in the travis cache until we have a working release on hackage.  Shouldn't be long: https://github.com/Soostone/uri-bytestring/pull/4

fhartwig: Ok, we'll leave it for now.

PR #68, opened 2015-05-07
hw service: store service session token in cookie when returning from…

fhartwig:
… thentos login

### DIFF
--- a/exec/BuildDocs.hs
+++ b/exec/BuildDocs.hs
@@ -177,8 +177,8 @@ instance ToSample (UserId, Timeout) where
 instance ToSample (UserId, ServiceId) where
     toSample = (,) <$> toSample <*> toSample
 
-instance ToSample ServiceSessionMetaData where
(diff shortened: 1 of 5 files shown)


fhartwig:
For some reason, the hw service currently doesn't notice when the user logs out of thentos. I suspect that this is a bug in thentos, though. Will investigate.

mf59816:
if you mean "hw doesn't notice if user logs out of thentos", then i can explain: thentos clears the DB on user logout, but it doesn't notify the service.  it doesn't even have a channel for that; hw would need to ask thentos if the service session token is still valid, but i haven't found that happening anywhere either.

Am I talking about the same thing as you?

fhartwig:
@mf59816 Yes, we're talking about the same thing. `verifyToken` is called by every request to the hw service, and it should return `False` if the session doesn't exist any more. But that doesn't seem to happen.

fhartwig:
update: service sessions aren't cleaned up when a thentos session is deleted. Fixing.

PR #71, opened 2015-05-12
Generate transaction boilerplate with template haskell

fhartwig:


mf59816:
Looks good.  Just a few details.

PR #82, opened 2015-05-19
selenium experiments

fhartwig:
Don't merge any of this yet.


fhartwig:
Seems to be working now (tests failing on travis are unrelated, as far as I can tell).
Also, probably a new record for fewest LOC/h.

PR #85, opened 2015-05-20
Travis dependency cache stuff

fhartwig:
Don't merge this yet, this pr is just there so I can see what travis does.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ before_install:
   - make start
   - cd ../..
 install:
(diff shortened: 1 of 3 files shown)


fhartwig:
We're now using a cabal sandbox to cache the dependencies, so this now works even if we have dependencies that need to be `cabal sandbox add-source`'d.

mf59816:
```
Warning: The package list for the local repo

'/home/travis/build/liqd/thentos/.cabal-sandbox/packages' is missing. The repo

is invalid.
```

PR #95, opened 2015-05-22
Fix hello-world service

fhartwig:


fhartwig:
I constructed the request body outside of `makeRequest` to keep `makeRequest` more general (so it can be used for api calls whose body is not a single `ServiceSessionToken`), but since we don't actually need that, your way is probably better.

mf59816:
that's my opinionated opinion anyway.  (-:

a generalization with one instance is usually a bad smell...

PR #114, opened 2015-05-28
`Test.WebDriver.Missing` module for more robust page loads.

### LINKED ISSUE #111: webdriver non-determinism.
I think one problem that we have (perhaps it's the only one?) is that `findElem` sometimes fails because the page is not ready yet.  Two possible solutions (I think):

1. Add something like `waitForPageReady` to webdriver (or make it a `Missing` module for now).  This probably requires javascript hacking over webdriver.  Unpleasant!
2. Add something like `retryCall` that takes a timeout and a frequency in ms and a `WD`, and calls the `WD` in the frequency until timeout is reached.  Easier to implement (I think), but ugly!

mf59816:
Fixes #111 

Still testing.  DO NOT MERGE.

### DIFF
--- a/tests/Test/WebDriver/Missing.hs
+++ b/tests/Test/WebDriver/Missing.hs
@@ -1,0 +1,71 @@
+{-# LANGUAGE FlexibleContexts     #-}
+{-# LANGUAGE KindSignatures       #-}
+{-# LANGUAGE LambdaCase           #-}
+{-# LANGUAGE OverloadedStrings    #-}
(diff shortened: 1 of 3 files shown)


mf59816:
So...  it's still possible to crash this if you run run `cabal test` in two different shells concurrently, but that is not supposed to work anyway (at least the tcp port listeners will be in each other's way).

Otherwise, I am theoretically and practically happy with this fix.

PR #122, opened 2015-06-02
Thentos.Backend.Api.Adhocracy3

mf59816:
A while ago I wrote a rest api that tries to act like adhocracy3.  I just un-bit-rotted it a little, but there is still one failing test case.

Test case should be fixed before merging.

### DIFF
--- a/src/Thentos/Backend/Api/Adhocracy3.hs
+++ b/src/Thentos/Backend/Api/Adhocracy3.hs
@@ -20,17 +20,9 @@
 -- git@github.com:liqd/adhocracy3.git:/docs/source/api/authentication_api.rst
 module Thentos.Backend.Api.Adhocracy3 where
 
-{-
-
 import Control.Applicative ((<$>), (<*>), pure)
-import Control.Concurrent.MVar (MVar)
 import Control.Monad.Except (throwError)
-import Control.Monad.IO.Class (liftIO)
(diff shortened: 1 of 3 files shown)


ChristianSi:
I fixed an unrelated typo and made the failing test case work -- though I can't say exactly what was the original problem. Instead of creating the request body via `cs . unlines` I rewrite it to use multi-line string syntax. That made the JSON parser happy, though I don't understand why it didn't accept the original version -- I sent it through an external JSON parser and it was parsed correctly.

Anyway, the rewritten multi-line string is shorter and not less readable than the original version, so I guess the change is no loss.

ChristianSi:
Assigned this back to @mf59816 -- you can merge it or give it back to me if you're still unhappy with something.

mf59816:
Most weird!  I wonder if we try and isolate this problem and open an aeson issue on it?

Anyway, the multi-line syntax looks good, too.  Also, I don't know why I had a string there in the first place, since there is a much better way to construct aeson values with combinators, and I just did that.

If you don't prefer the multi-line string version for some reason, and if travis is happy, you can merge this.

ChristianSi:
Merged. The problem with the original input was that there is a newline after the final `}` and Aeson doesn't seem to like that, reporting "invalid request body: endOfInput" instead. That looks very much like a bug so I'll try to reproduce it directly in Aeson and open a ticket.

PR #124, opened 2015-06-02
Make forms CSRF-proof

### LINKED ISSUE #115: CSRF security

fhartwig:
Fixes #115 

I'm not sure if we want to keep the dependency on snap-extras, because it pulls in several additional large dependencies (like haskell-src-exts) which we don't use at all. We could copy `blanketCSRF` and `handleCSRF` from https://github.com/ozataman/snap-extras/blob/master/src/Snap/Extras/CSRF.hs instead, they're fairly small and simple. What do you think?


mf59816:
Yes, please copy the functions to our `Snap.Missing` module.  Or rather make a `Snap.Inline` module, which would be less of a lie, and less annoying for the snap people if they every stumble over this.  (-:

mf59816:
how hard would it be to write a test or two?  (one for happy flow and one that gives evidence for security?)

fhartwig:
The happy flow is already tested by the tests that submit forms. I'll try to hack something up for the problematic case.

fhartwig:
Ok, unless I've missed something, all your points should be addressed.

PR #128, opened 2015-06-04
Switch from crypto-random to cryptonite package

### LINKED ISSUE #125: Upgrade to cryptonite
http://tab.snarc.org/posts/haskell/2015-06-02-announcing_cryptonite.html

we depend on some of the previously separate packages.  should be simple enough to un-depend on them and depend on cryptonite.

fhartwig:
Fixes #125

We might also want to consider using cryptonite's [scrypt functionality](http://hackage.haskell.org/package/cryptonite-0.2/docs/Crypto-KDF-Scrypt.html), even though the scrypt package we're currently using is not one of the packages that cryptonite replaces. I don't have a strong opinion either way, but I'd probably leave it for now.


fhartwig:
@mf59816 Fair enough. I was assuming that we still had transitive dependencies on crypto-random (although I'm not sure how to best check this), so I didn't take it out. I'm generally not sure how well our approach of manually maintaining the cabal.config works with transitive dependencies, which we are likely to miss.

mf59816:
well, the point of cryptonite is that all the other libraries don't have to be maintained any more, so if we still depend on them, something is wrong.

if you just run `cabal freeze` again on a fresh sandbox, then `git diff` will tell you what changed, and usually you should not need to change anything there.

PR #135, opened 2015-06-09
Better cache control for Snap

### LINKED ISSUE #112: Add cache-control headers for snap and servant.
The behavior should be "never cache anything", and there should be one place for snap and one for servant where this is done.

Once we run into performance issues, we can optimize based on the profiling results.

ChristianSi:
Define unknownPath handler and register it as last route.
Thus cache-control headers are also added to 404 Not Found error pages.

disableCaching: only add headers to GET/HEAD requests.
Also document which requests are cached by default.

This finally fixes #112 completely (I hope).


mf59816:
otherwise looks good.  @ChristianSi, you can merge if you have handled my two minor comments.

PR #137, opened 2015-06-10
Fix proxy module

### LINKED ISSUE #123: `Backend.Api.Proxy` is bit-rotten
fix it and document it better.  the purpose of this module is to provide a single end-point handler that proxies all requests under that end-point.  it should be the choice of the module user whether the prefix (the part of the path up to the end-point that calls the proxy) is removed or left intact.

fhartwig:
Fixes #123

### DIFF
--- a/src/Thentos.hs
+++ b/src/Thentos.hs
@@ -90,7 +90,7 @@ main =
                 void $ concurrently backend frontend
 
             RunA3 -> do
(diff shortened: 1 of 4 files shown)


mf59816:
A test would have been cool, but we will have to do that anyway for A3 soon.

Looks good!

PR #141, opened 2015-06-12
Hide not-found errors for secret objects.

### LINKED ISSUE #131: Catch authorization errors from inside actions.
The following code leaks the information whether a thentos session token is active, but allowing the attacker to distinguish between "access denied" and "False" responses:

```haskell
```

This is a general issue we need to solve: Authorization checks happen in the execution of `Action`, but some actions want to translate authorization errors to, e.g., "not found" responses.  We need a design for that.

mf59816:
Fixes #131.

For the simple example of looking up / testing for existence of `ThentosSession`s, I implement a lio policy that throws a `not found` exception both in the event that the lookup token comes up empty, and the event that it comes up with a session we have no clearance for.

The rest of the policy for `Thentos.Action` is still mostly missing.


ChristianSi:
Looks good except for some little note.

PR #143, opened 2015-06-13
Polymorphic Transactions, Actions

mf59816:
Work on #75.

This PR does NOT solve the issue, but:

- it makes writing and composing polymorphic transactions easier,
- it adds test cases that explain how `AsDB` works, and
- it cleans up the code a little.

Can be reviewed and possibly merged.

PR #144, opened 2015-06-13
Run hlint on travis

mf59816:
(I have opened #141, #143, and this PR on top of each other.  I wonder if that means that they can be merged in any order?)


mf59816:
The errors in module `Adhocracy3` can be ignored (will be fixed by @ChristianSi elsewhere).

ChristianSi:
Looks good. Problem is that the hlint errors in `Adhocracy3.hs` now temporarily break Travis but I hope to fix that with another PR today.

PR #145, opened 2015-06-15
Clean up Thentos.Action, Thentos.Backend.Api.Simple

mf59816:


mf59816:
I took out some work in progress and will PR that separately.  The commits above should be non-controversial enough, and I would like to get them out of the way soon.

ChristianSi:
I didn't run the tests, but the code looks good. Merging.

PR #147, opened 2015-06-15
Use email-validate package for UserEmail

ChristianSi:


exec/BuildDocs.hs
```diff
@@ -157,15 +157,17 @@ instance ToSample [ThentosSessionToken] [ThentosSessionToken] where
     toSample _ = Just ["abde1234llkjh", "47202sdfsg"]
 
 instance ToSample UserFormData UserFormData where
-    toSample _ = Just $ UserFormData (UserName "Kurt Cobain")
-                                   (UserPass "Hunter2")
-                                   (UserEmail "cobain@nirvana.com")
+    toSample _ = Just $ UserFormData (UserName "Kurt Cobain") (UserPass "Hunter2") sampleEmail
```
mf59816: I would prefer:

```haskell
  toSample _ = do
    sampleEmail <- toSample
    Just $ UserFormData (UserName "Kurt Cobain") (UserPass "Hunter2") sampleEmail
```

...  and then make your `sampleEmail` below a `ToSample` instance.

mf59816:
so...  travis scripts can handle missing sandbox cache file, but there is still a problem with the cabal dependencies and current hackage.  i think we need more constraints in the cabal file?

ChristianSi:
I made some little improvements as per Matthias' suggestion. It's now ready to merge, I think.

PR #153, opened 2015-06-16
Build hlint in cache.

mf59816:


mf59816:
I have re-built the sandbox cache with the new script.  let's see what travis does...

PR #162, opened 2015-06-22
A3: print user paths as full URLs

ChristianSi:
That's how A3 expects it.

(Sorry for the branch name, this had intervened when I had started out to do something else.)


src/Thentos/Backend/Api/Adhocracy3.hs
```diff
@@ -324,32 +324,39 @@ sendUserConfirmationMail smtpConfig user callbackUrl =
...
+userIdToPath :: ThentosConfig -> UserId -> Path
+userIdToPath config (UserId i) = Path $ domain <> userpath
+  where
+    domain   = cs $ exposeUrl beHttp
+    userpath = cs (printf "principals/users/%7.7i" i :: String)
+    beHttp   = case config >>. (Proxy :: Proxy '["backend"]) of
+                    Nothing -> error "userIdToPath: backend not configured!"
+                    Just v -> Tagged v
 
 userIdFromPath :: Path -> AC.Action DB UserId
```
mf59816: This signature could be generalized:

```haskell
 userIdFromPath :: MonadError m => Path -> m UserId
```

Then the tests could instantiate `m` with `Maybe` instead of `Action` and would be slightly easier to write.

ChristianSi:
@mf59816 Thanks for your suggestions, I've now handled all of them (I think).

PR #163, opened 2015-06-22
Implement parts of A3-SSO

fhartwig:


src/Thentos/Backend/Api/Adhocracy3Sso.hs
```diff
@@ -175,46 +176,45 @@ githubKey = OAuth2 { oauthClientId = "c4c9355b9ea698f622ba"
...
+    return . AuthRequest . cs $ authorizationUrl githubKey `appendQueryParam` [("state", cs $ fromSsoToken state)]
 
 
 -- | FIXME: document!
 githubConfirm :: ST -> ST -> AC.Action DB A3.RequestResult
 githubConfirm state code = do
-
-    -- FIXME: lookup state in DB and crash if it does not exist.  remove if it does exist.
-
+    ssoTokenExists <- AC.update'P $ LookupAndRemoveSsoToken (SsoToken state)
```
mf59816: There should be no transactions in `Backend.*`, since they have no authorization checks.

(Of course for the same reason, there should also be no `import LIO.TCB` for the same reason.  @fhartwig, decide whether which of them should be fixed now or later.)

mf59816:
Everything here is between acceptable and good (except for the whitespace :-).

Please review comments, fix to your liking and pass back to me.

src/Thentos/Backend/Api/Adhocracy3Sso.hs
fhartwig: Ok. I put it here because I thought it was too special-purpose to go into `Thentos.Action`, but can move it. Not sure about the LIO stuff, will look into it.

PR #164, opened 2015-06-23
Configurable custom header names

ChristianSi:
Allows redefining the names used for custom headers.

This is used in the A3 proxy to read and write the headers which A3 understands, e.g. "X-User-Path" instead of "X-Thentos-User".

### DIFF
--- a/src/Thentos/Backend/Api/Adhocracy3.hs
+++ b/src/Thentos/Backend/Api/Adhocracy3.hs
@@ -295,7 +295,8 @@ thentosApi actionState = enter (enterAction actionState Nothing) $
 api :: AC.ActionState DB -> Server Api
 api actionState =
        thentosApi actionState
-  :<|> serviceProxy actionState
+       -- TODO define distinct function
+  :<|> serviceProxy renderThentosHeaderName actionState
 
 
 -- * handler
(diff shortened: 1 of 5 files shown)


src/Thentos/Backend/Api/Adhocracy3.hs
```diff
@@ -345,6 +346,12 @@ login r = AC.logIfError'P $ do
 
 -- * aux
 
+-- | Render Thentos/A3-specific custom headers using the names expected by A3.
+renderA3HeaderName :: RenderHeaderFun
```
mf59816: Shouldn't it be `renderA3HeaderNames`?

ChristianSi: The other function is called `renderThentosHeaderName`, I just adapted that name. Since either function alway generates a single header name, I guess it's OK.

mf59816:
Travis has a point there.  It's not immediately obvious to me what needs to be changed.

src/Thentos/Backend/Api/Adhocracy3.hs
mf59816: oh, you're right.

PR #167, opened 2015-06-24
Document some of our development practices

ChristianSi:


ChristianSi:
A (somewhat) unrelated note re the docs. In the styleguide there is a

> FIXME: `LT` for `Data.Text.Lazy` is `less than` from Prelude.  Is there a better name?

But LT in Prelude is a constructor, while we use LT as name for a type and a module, hence there are no conflicts. Therefore, why not delete the FIXME?

mf59816:
Re `LT`: good point.  could you delete the fixme?  thanks (:

ChristianSi:
I think I've addressed all issues.

PR #171, opened 2015-06-26
Styleguide: test module names.

mf59816:
For discussion.


mf59816:
@ChristianSi: please pass on if you agree, or return to me if you object.

ChristianSi:
Looks reasonable to me :+1:

Issue #110, opened 2015-05-28
A3 interoperability


mf59816:
This is a big ticket, and will probably require more explanation from my part, but it's a start.

Each of the following steps should be self-contained.  Changing adhocracy code will not be necessary until we reach step 3.

1. run thentos as a trivial rest proxy for adhocracy (trivial == "does not change behavior, only listens in.")
2. catch user registration requests, and feed thentos db with users.
3. make adhocracy (1) accept session tokens without checking, and (2) fill its user resources via rest api calls to thentos.
4. sub-services: different instances on the same adhocracy installation should be different services for thentos, and get different pseudonyms for the same user.

the last point certainly requires more thinking and should probably be considered an outlook rather than part of this ticket.

we may also want to start opening sub-tickets on this.

mf59816:
opened https://github.com/liqd/adhocracy3.mercator/issues/1349

fisx:
ChristianS: giving you the lead on this.

(Item 4. from the list above should become a new ticket once this is done.  We may want to do SSO first, too.)

ChristianSi:
I have Thentos running as Proxy between A3 front- and backend, but there are still various problems to solve until anything useful will be possible. E.g.

1. The proxy refuses requests if the user is not logged in ("ActionErrorThentos NoSuchThentosSession"). But of course, such requests are necessary for user creation and user login as well as for anonymous browsing!

2. Our proxy doesn't understand the A3 batch API (batch request). Sooner or later we may have to tackle them, and that's not a small task.

I'll work on (1) first.

ChristianSi:
The happy flow is working now. I don't think we'll need batch requests (at least for the time being), since user creation, account activation and user login are never sent in batch requests by the A3 frontend.

I'm looking at the error cases now -- it may be possible to mimic the typical A3 backend error messages so that the A3 frontend doesn't need to be changed.

ChristianSi:
I believe that with the completion of #168 this task is complete. Except for (4), sub-services or personas, which I would see as a different task, especially as the requirements for that one are still very unclear.

@fisx If you agree, please close this ticket.

fisx:
@ChristianSi: agreed (:

PR #182, opened 2015-06-30
Proxy module: allow anonymous requests and a default proxy

ChristianSi:
* Add support for forwarding requests from users that aren't logged in.
* If no X-Thentos-Service is specified, we use a default proxied app, if  configured (new section in the config).
* Add sample config for running A3 with Thentos in proxy mode.

Part of the ongoing work on #110.

### DIFF
--- a/devel.config-a3-sample
+++ b/devel.config-a3-sample
@@ -1,0 +1,35 @@
+# Sample config for running A3 with Thentos in proxy mode.
+# Suggested ports: A3 frontend on 6551, Thentos proxy on 6546, A3 backend on 6541.
+
+command: "runA3"
+
+backend:
+    bind_port: 6546
+    bind_host: "127.0.0.1"
+
+#frontend:
+#    bind_port: 7002
+#    bind_host: "127.0.0.1"
+
+proxy:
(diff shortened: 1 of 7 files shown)


ChristianSi:
Pushed some more commits since @mf59816 wasn't yet on this. User creation and activation in proxy mode now almost works, with two caveats:

1. The frontend doesn't understand the answers sent by the Thentos proxy. I'll investigate that.

2. Since users are only created in Thentos but not also in A3, future actions with logged-in users won't be possible. We'll probably have to modify the proxy to duplicate user creation in A3 as well as Thentos.

fisx:
On '2.' in your last comment: I was thinking the handlers should not only create the user in the thentos database, but also in the backend, via the rest api that is supposed to be used by the frontend.  Two advantages:

1. this is more convenient if the a3 backend can be taught to let the proxy create users without email authentication when in proxy mode.
2. if we want to keep some user information to ourselves, it is easier to do that in the handler than by filtering the actual frontend requests.

ChristianSi:
I addressed @mf59816's comments (more or less). @mf59816, want to take another look and then merge?

fisx:
(did the tests locally)

PR #184, opened 2015-07-02
poly transactions

### LINKED ISSUE #75: Polymorphism over DB
#74 introduces polymorphism over the `DB` type for transactions and actions.  But there are several places where the type gets materialized back into `DB` where we don't want that.

The culprit is `makeAcidic`.  It takes a concrete type `DB`, and not a type variable  After the transactions have thus been frozen back into monomorphicness, actions cannot unfreeze them.  Consider module `Thentos.Action.Core`: `update'P` and `query'P` are not the problem, they don't even need the constraint on the polymorphic `db` variable.  But once the actual frozen transaction is passed to it, the action type is frozen as well.  And everything becomes `DB` again.

I wrote this during a well-typed review telco on 2015-04-15:
 
    then how do we abstract over actions?  `makeThentosAcidic` (or whatever it is called)
    could construct a type class and instance that contains all the transactions it constructs,
    and somebody else could come and overload it.  then actions could use the class instead
    of the concrete types.

This may work, but obviously needs more thinking.

fhartwig:
Fixes #75

### DIFF
--- a/src/Thentos/Transaction/TH.hs
+++ b/src/Thentos/Transaction/TH.hs
@@ -53,14 +53,18 @@ dbEvents =
     ..
     ]
 
+-- for db types that extend an existing db type, dbEvents will look like this:
+dbEvents :: (db `Extends` DB, db `Extends` CustomDB => [Event db]
+dbEvents = [...new events...] ++ ParentDb.dbEvents
+
 instance IsAcidic DB where
     acidEvents = dbEvents
 -}
 
-makeThentosAcidicPhase2 :: Name -> [Name] -> Q [Dec]
-makeThentosAcidicPhase2 stateName transNames = do
+makeThentosAcidicPhase2 :: Name -> [Name] -> [Name] -> [Name] -> Q [Dec]
+makeThentosAcidicPhase2 stateName transNames inheritedDbTypes inheritedEvents = do
     decs <- concat <$> mapM (acidifyTrans stateName) eventNames
-    eventList <- mkEventList stateName eventNames
+    eventList <- mkEventList inheritedDbTypes inheritedEvents stateName eventNames
     isAcidic <- makeIsAcidic stateName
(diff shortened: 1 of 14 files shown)


fisx:
@fhartwig: please fix travis.  you can disable the entire module if you can't think of anything better.  (i couldn't.)

fhartwig:
@fisx I've disabled warnings on parse errors. If the syntax is actually incorrect, the build will fail anyway, so I don't really see the point of leaving it on.

src/Thentos/Transaction/TH.hs
```diff
@@ -7,36 +10,243 @@ module Thentos.Transaction.TH
...
-makeThentosAcidicPhase2 stateName eventNames =
-    makeAcidic stateName $ map dropPrefix eventNames
+makeThentosAcidicPhase1 :: Name -> [Name] -> Q [Dec]
+makeThentosAcidicPhase1 dbTypeName names =
+    concat <$> mapM (processTransaction $ ConT dbTypeName) names
 
-processTransaction :: Name -> Q [Dec]
-processTransaction functionName = do
+{-
+-- for every transaction, we need the following:
```
fisx: I think documentation for this module is more relevant than for others.  `docs/concepts/AcidPoly.lhs` is a big help, but it is not very specific on the `TH` api.

I would like this module to have a haddock section, that links to the concepts and then elaborates on the `TH` specifics, and answers questions like: How do you call which function if you want to use this as a library?  (Or if you want to extend the kernel?)

fisx:
@fhartwig: please review comments and commits.

src/Thentos/Transaction/TH.hs
fisx: (i'll open an issue for this.  can be done separately, after the merge.)

fhartwig: (Punted to a separate pr)

PR #187, opened 2015-07-07
Create users in the A3 backend as well as in A3

ChristianSi:


src/Thentos/Backend/Core.hs
```diff
@@ -101,6 +101,8 @@ actionErrorToServantErr e = do
         { errBody = "error accessing user info" }
     _thentos (SsoErrorCouldNotGetAccessToken _) = pure $ err500
         { errBody = "error retrieving access token" }
+    _thentos (A3BackendError msg) = pure $ err500
+        { errBody = "error talking to A3 backend: " <> cs msg }
```
fisx: I just pushed 6533ee7 regarding this.

A3BackendError is very untyped.  There is no strong reason to not do that besides that it's not necessary and not idiomatic.  I would use different types `A3BackendBadResponseCode`, `A3BackendNoSuchUser`, ..., and make each carry the details in typed form.  See 6533ee7 for more explanation.

src/Thentos/Transaction/Transactions.hs
```diff
@@ -49,6 +49,10 @@ assertUser mUid user = polyQuery $ ask >>= \ db ->
        | userFacetExists (^. userEmail) mUid user db -> throwT UserEmailAlreadyExists
        | True -> return ()
 
+-- | Assert that no user with the same name or email address already exists in the DB.
+trans_assertUserIsNew :: (AsDB db) => User -> ThentosQuery db ()
+trans_assertUserIsNew = assertUser Nothing
```
fisx: see earlier comment on the missing exception in `addUnconfirmedUserWithId`.  if you fix that, this here may become unnecessary.  in that case i'd like to keep the api simple.  perhaps `lookupUser` can cover for this one?

src/Thentos/Transaction/Transactions.hs
```diff
@@ -530,7 +543,8 @@ $(deriveSafeCopy 0 'base ''UpdateUserFieldOp)
 
 transaction_names :: [Name]
 transaction_names =
-    [ 'trans_allUserIds
```
fisx: thank you.  (-:  i've been meaning to do that for a while now.

src/Thentos/Transaction/Transactions.hs
fisx: @ChristianSi has pointed out that this has nothing to do with `UserId`, but checks freshness of name and email.  Please ignore previous comment.

ChristianSi:
As a side effect, my latest commit fixes #190.

ChristianSi:
@fisx I think it makes sense to merge this now so as not to interfere with the big refactoring. Some of your comments are still unaddressed, but I'll take care of them in a different PR.

fisx:
looks good.  i will unassign me; @ChristianSi, if you find that travis is happy, you can merge.  if you don't, i'll do it tonight.

PR #191, opened 2015-07-08
Make actions polymorphic in db type where possible

fhartwig:


fhartwig:
Not sure what travis's issue is with my first commit, this builds fine locally. Also, the line it points to is a comment? Weird. Let's hope it's just some travis bug, if this keeps failing you should test it locally.

fisx:
I agree.  From the way travis fails, it is clear that it doesn't attempt to build the code from this branch.

fisx:
Oops, there was a bug here, but it only shows for a fresh build, not for an incremental bulid.  It gets weirder and weirder...

The bug: `prepareUserData` is monomorphic and forces monomorphicity on `addUnconfirmedUser`.  Will fix elsewhere now.

PR #195, opened 2015-07-09
Refactor: multiple packages.

### LINKED ISSUE #193: Polymorphic Error handlers

```haskell
```

Further reading:

fisx:
Introduce 3 packages: thentos-core, thentos-tests, thentos-adhocracy.

thentos-core contains no tests: the test infrastructure should be available to all using packages during development, but not in production mode.  in order to achieve that, thentos-tests provides the infrastructure a a library (similar to hspec, quickcheck, etc.).  in order to avoid cyclic dependencies, the tests for thentos-core have been migrated to thentos-tests.

in many places, the original code was not sufficiently polymorphic.  in fact, it must be true that this fixes #193.


fisx:
Not ready for merge yet!

thentos-adhocracy/src/Thentos/Adhocracy3/Types.hs
```diff
@@ -0,0 +1,103 @@
...
+
+-- FIXME: workaround implementation until #193 is fixed:
+
+data ThentosError' =
+      A3BackendErrorResponse Int LBS
+    | A3BackendInvalidJson String
+  deriving (Eq, Show, Typeable)
+
+throwA3Error :: ThentosError' -> a
+throwA3Error = error . show
```
jkarni: Why not `MonadError` (like `throwCoreError`)?

thentos-adhocracy/tests/Thentos/Adhocracy3/Backend/Api/SimpleSpec.hs
```diff
@@ -158,7 +162,7 @@ spec =
 --                  return ()
 
         -- FIXME currently not working because of Servant quirks on failures
-        --describe "create user errors" . before (setupTestBackend RunA3)
+        --describe "create user errors" . before setupTestBackend
```
jkarni: If we use HEAD this should work.

thentos-tests/src/Thentos/Test/Core.hs
```diff
@@ -139,15 +139,15 @@ teardownBare (TS tcfg) = do
 
 -- * DBTS
 
-setupDB :: IO DBTS
+setupDB :: forall db . (db `Extends` DB, IsAcidic db, Eq (ThentosError db)) => IO (DBTS db)
```
jkarni: Where can I read about `Extends`?

fhartwig: There is an explanation at `https://github.com/liqd/thentos/blob/master/docs/concepts/AcidPoly.lhs`

thentos-tests/src/Thentos/Test/Core.hs
```diff
@@ -208,14 +212,19 @@ setupTestServerFull = do
...
 teardownTestServerFull (FTS tcfg db backend _ frontend _ _) = do
     cancel backend
     cancel frontend
     teardownDB $ DBTS tcfg db
 
 
-loginAsGod :: ActionState DB -> IO (ThentosSessionToken, [Header])
+loginAsGod :: forall db .
+        ( db `Extends` DB , IsAcidic db
+        , db ~ DB  -- FIXME https://github.com/liqd/thentos/issues/193
```
jkarni: If we can't get rid of this constraint yet, shouldn't be just simplify to `db` to `DB`, and then change back when it works?

fisx: We could, but I wanted to write down as much of the *intended* structure as possible, and I also want to fix this next, so this line should not stay there for long.  So I'd like to leave it as is.

thentos-adhocracy/src/Thentos/Adhocracy3/Types.hs
fisx: true.  fixing.

thentos-adhocracy/src/Thentos/Adhocracy3/Types.hs
```diff
@@ -0,0 +1,103 @@
...
+
+throwCoreError :: (e ~ ThentosError DB, MonadError e m) => e -> m a
```
jkarni: `throwCoreError :: (MonadError (ThentosError DB) m) => ThentosError DB -> m a` doesn't work?

thentos-adhocracy/src/Thentos/Adhocracy3/Types.hs
fisx: correction: the type *should* be

```haskell
throwA3Error :: (e ~ ThentosError', MonadError e m) => e -> m a
```

but that does not type.  it will type once #193 has been fixed, so i'd like to leave this as is for now.

thentos-tests/tests/Thentos/ActionSpec.hs
```diff
@@ -173,7 +174,7 @@ spec_session = describe "session" $ do
             return ()
 
     describe "lookupThentosSession" $ do
-        it "works" $ \ (DBTS _ astate :: DBTS) -> do
+        it "works" $ \ (DBTS _ astate :: (DBTS DB)) -> do
```
jkarni: Not the world's most descriptive test name ;)

fisx:
two open issues:

1. travis
2. #193 

i would like to merge this anyway and fix them separately.

thentos-core/Makefile
```diff
@@ -0,0 +1,38 @@
+SHELL=/bin/bash
+HLINT=hlint
+
+test:
+	cabal test --test-options="--skip selenium"
+
+test-all:
+	cabal test
+
```
fhartwig: `test` and `test-all` are obsolete here.

PR #198, opened 2015-07-13
extensible errors (variant 1)

### LINKED ISSUE #193: Polymorphic Error handlers

fisx:
Fixes #193 

I'll write down a variant suggested by @jkarni now so we can compare.

### DIFF
--- a/thentos-adhocracy/src/Thentos/Adhocracy3/Types.hs
+++ b/thentos-adhocracy/src/Thentos/Adhocracy3/Types.hs
@@ -72,7 +72,7 @@ instance EmptyDB DB where
 
 instance DB `Core.Extends` Core.DB where
     focus f (DB db) = DB <$> f db
-    asDBThentosError = ThentosA3ErrorCore
+    thentosErrorFromParent = ThentosA3ErrorCore
 
 data instance Core.ThentosError DB =
       ThentosA3ErrorCore (Core.ThentosError (Core.DB))
--- a/thentos-adhocracy/src/Thentos/Adhocracy3/Types.hs
+++ b/thentos-adhocracy/src/Thentos/Adhocracy3/Types.hs
@@ -96,6 +96,15 @@ throwA3Error = throwError
 throwCoreError :: (e ~ ThentosError Core.DB, e' ~ ThentosError DB, MonadError e' m) => e -> m a
 throwCoreError = throwError . ThentosA3ErrorCore
 
+instance ThentosErrorToServantErr DB where
+    thentosErrorToServantErr (ThentosA3ErrorCore e) = thentosErrorToServantErr e
+    thentosErrorToServantErr e@(A3BackendErrorResponse code msg) = do
+        logger ERROR (show e) >> err500 { errBody = "exception in a3 backend" }
+
+  thentosErrorToServantErr e@(A3BackendInvalidJson json) = do
+        logger ERROR (show e) >> err500 { errBody = "exception in a3 backend: received bad json" }
+
+
 makeLenses ''DB
 
 $(deriveSafeCopy 0 'base ''DB)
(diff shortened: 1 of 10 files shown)


fisx:
I think this is a sufficiently trivial addition to master, and it is sufficiently useful (it closes #193!) that it should be merged now.  `2015-07-13-mf-poly-errors-2nd` should be considered, but later and independently.

@jkarni, please could you take a quick look and then either reject or merge?

thentos-adhocracy/src/Thentos/Adhocracy3/Types.hs
```diff
@@ -72,7 +72,7 @@ instance EmptyDB DB where
 
 instance DB `Core.Extends` Core.DB where
     focus f (DB db) = DB <$> f db
-    asDBThentosError = ThentosA3ErrorCore
+    thentosErrorFromParent = ThentosA3ErrorCore
```
jkarni: I tend to dislike comments and FIXMEs etc., since it usually makes it feel like it is okay to only half do a task or half implement a feature. And commented out code bitrots quickly (it is never typechecked), and soon becomes more confusing than helpful. But again, this is a very general point, and not specific to this PR.

thentos-core/src/Thentos/Types.hs
```diff
@@ -89,14 +89,18 @@ class ( Typeable dbParent, Typeable dbChild
     focus :: Lens' dbChild dbParent
     -- ^ Apply anything that is intended for @dbParent@ to @dbChild@.
 
-    asDBThentosError :: ThentosError dbParent -> ThentosError dbChild
+    thentosErrorFromParent :: ThentosError dbParent -> ThentosError dbChild
     -- ^ If a transaction or action associated with 'dbParent' throws an error, use
     -- this function to convert it to an error that can be thrown by
     -- transactions or actions associated with @dbChild@.
 
+    thentosErrorToParent :: ThentosError dbChild -> Maybe (ThentosError dbParent)
```
jkarni: This seems very strange. Basically we've managed to allow functions that shouldn't be handling certain kinds of errors to do so, by mapping those errors to `Nothing`. That is, in my opinion, a lot of type safety lost.

jkarni:
I have general objections, which are only somewhat related to the specifics of the PR. I think the fact that TH made it hard to implement a simple change in the types is evidence against the TH approach (simple changes being simple is the acid test of maintainability). I think polymorphic DBs are making everything increasingly complex (and possibly also have against them the fact that TH is needed). And I think continuing down this route because this change is in fact fairly small is a sunk cost fallacy. But ultimately, if those are the choices we're going with (and if we're okay with the loss of type safety I mentioned above, which is a more concrete issue) this PR itself looks good to me.

thentos-adhocracy/src/Thentos/Adhocracy3/Types.hs
fisx: I agree.  Which error or FIXME did trigger this comment?  (Can't find anything in the code around it.)

thentos-core/src/Thentos/Types.hs
fisx: Good point.  The way we use this function does not have this problem, but other ways in which we *could* use it do.  I need to think about this.

fisx:
`TH` is not needed for polymorphic acid-state, but for acid-state, period.

I'm not arguing we should continue with the current approach because I think it's expensive to change the approach, but because I like the approach.  I don't think it's as complicated as you make it look either.

Perhaps we can do something to make the `TH` code more maintainable?

thentos-adhocracy/src/Thentos/Adhocracy3/Types.hs
jkarni: I just mean maintaining rather than removing commented out code.

fisx: Ah.  I totally missed that.  Thanks.

fisx:
Tests pass locally.  I addressed all issues except:

1. `thentosErrorToParent` (see #207)
2. polymorphic acid-states (this is way too fundamental a change for this PR.)

fisx:
(@jkarni: please merge if you have no further objections.)

fisx:
closing in favour of #211.

PR #202, opened 2015-07-13
Switch to servant master

ChristianSi:
First commit is an attempt to make a servant POST endpoint that responds with 200.

Doesn't work yet, because `methodRouter` only exists on master.


jkarni:
Compiles, but don't merge yet - I'd like to run the tests first, and document how to use the submodule.

thentos-core/src/Thentos/Backend/Core.hs
```diff
@@ -175,10 +187,10 @@ instance (HasServer subserver) => HasServer (ThentosAssertHeaders :> subserver)
...
     type ServerT (ThentosAssertHeaders :> subserver) m = ServerT subserver m
 
-    route Proxy subserver request respond = case badHeaders $ requestHeaders request of
-        []  -> route (Proxy :: Proxy subserver) subserver request respond
-        bad -> respond . RR . Right . responseServantErr  -- FIXME: use 'left' instead of all this?  yields a type error, though.
-             $ err400 { errBody = cs $ "Unknown thentos header fields: " ++ show bad }
+    route Proxy subserver = WithRequest $ \ request -> route (Proxy :: Proxy subserver) $
+       case badHeaders $ requestHeaders request of
+          []  -> subserver
+          bad -> return $ failWith $ HttpError status400 (Just . cs $ "Unknown thentos header fields: " ++ show bad )
```
fisx: space before closing paren, but not after opening.

thentos-core/src/Thentos/Backend/Core.hs
```diff
@@ -120,6 +121,17 @@ actionErrorToServantErr e = do
     _permissions _ = logger DEBUG (ppShow e) >> pure (err401 { errBody = "unauthorized" })
 
 
+-- * custom servers for servant
+
+data Post200 (contentTypes :: [*]) a
+    deriving Typeable
+
+instance ( AllCTRender ctypes a ) => HasServer (Post200 ctypes a) where
+
```
fisx: (i don't like the above empty line, but there is nothing in the style guide to prevent you from keeping it.)

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "servant"]
+	path = servant
```
fisx: Should we place this in a directory `submodules`?

fisx: also, we now still need to call `cabal sandbox add-source` for this, right?  is there a way around that?  if not: it needs to be added to the docs.

ChristianSi:
What's the current status of this?

jkarni:
I made the suggested fixes, but installing with `make install2` runs into dependency issues (I think it might have been doing so before), and I'd like to have it working before merging anything.

fisx:
with cabal-install, i have the problem that the test suite's stdout only becomes visible after the test suite run has completed.  it seems like this is happening in the last travis run, and the test suite just takes more than 10 minutes.  perhaps stack has a way of showing test output incrementally, line by line?

fisx:
configifier 0.0.5 is out.  @jkarni, can you bump the dependency?

misc/thentos-install.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash -
+#===============================================================================
+#
+#          FILE: thentos-install.sh
+#
+#         USAGE: ./thentos-install.sh
```
fisx: the script wants to run from `misc/..`, so this usage info is a bit misleading.

fisx:
:+1:

ChristianSi:
I made a little line note, otherwise this looks OK to me. Well, it's already merged anyway...

The README should be updated in a different PR, since the cabal install stuff is partially outdated.

I'm still trying to install via `misc/thentos-install.sh`, no errors so far. I'll also try via stack as that seems the way to go in the future...

jkarni:
@ChristianSi I downgraded because it works with 1.18, and 1.18 is the default version of cabal that ships with 7.8.4 (including the default stack version).

jkarni:
@ChristianSi https://github.com/liqd/thentos/pull/220 has README changes

ChristianSi:
:+1:

ChristianSi:
Install script worked as it should :+1:

PR #203, opened 2015-07-13
Cleanup code redundancy in Thentos.hs and Thentos/Adhocracy3.hs

fisx:


ChristianSi:
Also, lots of redundant import warnings in `Thentos.Adhocracy3`.

I think this needs some more cleanup.

ChristianSi:
Looks good now.

PR #204, opened 2015-07-13
Refactor build-docs executables

fisx:


ChristianSi:
I get an error calling the executable:

    thentos-adhocracy$ build-docs 
    build-docs: ./docs/generated: createDirectory: does not exist (No such file or directory)

thentos-core/src/ThentosDocs.hs
```diff
@@ -0,0 +1,13 @@
+module ThentosDocs (makeMain) where
+
+import Data.Thyme.Time ()
```
jkarni: What does this import do?

fisx:
@ChristianSi: fixed, thanks.

thentos-core/src/ThentosDocs.hs
fisx:
```
$ git grep 'import Data.Thyme.Time ()' | grep required
thentos-adhocracy/src/Thentos/Adhocracy3/Types.hs:import Data.Thyme.Time () -- required for NominalDiffTime's num instance
thentos-core/src/Thentos/Types.hs:import Data.Thyme.Time () -- required for NominalDiffTime's num instance
```

It would be nice to be able to import instances explicitly as well...

jkarni: But where is this instance being used?

fisx: ah.  good question.  hm...  probably used to be used somewhere, but isn't any more?  i'll see if i can remove it, and PR elsewhere.

PR #205, opened 2015-07-13
Caseless command-name parsing

### LINKED ISSUE #176: make --command flag case invariant.
The parser for the `Command` type in `Thentos.Config` needs to:

- ignore case
- crash reliably if there are two constructors that are only distinguishable if case sensitively parsed.

ChristianSi:
Fixes #176.

There is now also a nice error message if the specified command is unknown.

I haven't specifically handled the "crash reliably if there are two constructors that are only distinguishable if case sensitively parsed" requirement, but I think if someone defines an enum

    data Command = Run | RuN | ...

that's pathological enough to not deserve special treatment.

The current implementation will simply select the first matching constructor in that case.

### DIFF
--- a/thentos-adhocracy/devel.config
+++ b/thentos-adhocracy/devel.config
@@ -5,7 +5,7 @@
 #       adhocracy.frontend.rest_url = http://localhost:6546
 # * Call bin/buildout
 
-command: "runA3"
+command: "run"
 
 backend:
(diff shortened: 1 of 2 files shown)


fisx:
LGTM.  please decide if you want to delete either or both of my commits and then merge.

PR #211, opened 2015-07-14
extensible errors (2nd attempt at variant 1)

### LINKED ISSUE #193: Polymorphic Error handlers

fisx:
Fixes #193; obsoletes #198.

See discussion there.  I hope this time I got it right.  Review should probably happen on the total changes, not individual commits.  (I'd keep the commits anyway in case we ever need to inspect intermediate states.)


jkarni:
LGTM (though if we care about keeping our commit history clean, it'd probably be nice to squash the commits).

@ChristianSi could you compile and run the tests? I don't have selenium set up yet...

fisx:
I've run the tests locally and successfully.  Of course I have claimed that before...  (-:

PR #214, opened 2015-07-14
Handle errors in Thentos.Smtp

### LINKED ISSUE #173: handle errors in Thentos.Smtp
if the sendmail executable is not installed, or if the email cannot be sent, sendMail must return an error, not throw it.  everybody calling it must handle that error reasonably.

fhartwig:
Fixes #173


fisx:
@fhartwig: "you are right" to both your responses.  re-assigning to you.

PR #217, opened 2015-07-15
Cleanup: bogus qualified imports; bogus parens.

fisx:


thentos-adhocracy/src/Thentos/Adhocracy3/Types.hs
```diff
@@ -44,36 +44,36 @@ newtype DB = DB { fromCoreDB :: Thentos.Types.DB }
   deriving (Eq, Show, Typeable, Generic)
 
 instance EmptyDB DB where
-    emptyDB = DB $ Thentos.Types.emptyDB
+    emptyDB = DB emptyDB
 
-instance DB `Thentos.Types.Extends` Thentos.Types.DB where
+instance DB `Extends` Thentos.Types.DB where
```
jkarni: Shouldn't the RHS also be just `DB`?

fisx: `DB` is the one from thentos-adhocracy; `Thentos.Core.DB` is the one from thentos-core.  The reflexive instance is below.

PR #237, opened 2015-07-20
Make user creation and activation work

ChristianSi:
This PR modifies our replies so that the A3 frontend is happy:

* POST responses have status code 200 instead of 201.
* Add and correctly fill "updated_resources" field to allow the A3 frontend to update its cache.
* Access-Control headers are added to allow embedding.

Login still doesn't work, which is kinda expected since the A3 backend checks the X-User-Token which our proxy deliberately doesn't pass on. So modifying A3 will be next.


fisx:
looks good; none of my comments strictly requires any action.  @ChristianSi: please review my comments and either assign back to me or merge.

ChristianSi:
I pushed another commit refactoring the CORS policy handling.

@fisx If you can simplify the `toSample` definition in `Thentos/Adhocracy3/Backend/Api/Docs/Simple.hs`, please do so. Otherwise I think it's ready for merge.

PR #240, opened 2015-07-21
Simplify instance FromJSON Prio by using readMay

ChristianSi:


fisx:
looks good, will merge later today.

Issue #224, opened 2015-07-17
Better proxying


jkarni:
Looking at https://github.com/liqd/thentos/blob/master/thentos-core/src/Thentos/Backend/Api/Proxy.hs#L49, it seems to me like our proxying isn't optimal. For one, it seems to me like we might get space leaks. Maybe it's worth investigating whether in fact this is the case, but we could also just use an existing soution (such as [this](https://hackage.haskell.org/package/http-reverse-proxy-0.4.2/docs/Network-HTTP-ReverseProxy.html#g:2) ).

jkarni:
We are also using `withManager`, and so creating a new manager on each request, which the documentation says is a bad idea.

fisx:
is this done?  (which pr?)

jkarni:
Done in #241

PR #241, opened 2015-07-22
Change proxy to use http-reverse-proxy.

jkarni:
Switch to global client manager (rather than creating a new one per
        request).
        Add tests for proxied requests.


jkarni:
Fixes #224.

fhartwig:
:+1: from me, I'm just waiting for travis to pass before merging

PR #248, opened 2015-07-22
Small steps for making A3 login work

ChristianSi:
Not there yet, but two little steps in the right direction:

* User tokens weren't serialized correctly.
* Looking up the session based on the X-User-Token header failed because of clearance issues.


ChristianSi:
@fisx @jkarni Can one of you merge this? My understanding is that @jkarni already reviewed it and had no complaints, but wanted to wait for Travis before merging.

jkarni:
done

PR #262, opened 2015-07-27
Werror and other travis details

jkarni:
Add `-Werror`, and generate a `cabal.config' that accords with the stack versions.


jkarni:
@fisx we're not using stack

PR #263, opened 2015-07-27
Autocreate services listed in the config

### LINKED ISSUE #247: Services should be auto-created when the config is read
Services have IDs which are currently generated dynamically via the Thentos frontend. However,as they must be configured in the config file, that means that a config file will only ever work for a single local setup, and only as long as the DB is not deleted.

To allow sharing config files between different users, services that don't exist in the local DB when a config is read should be auto-created. That means, to add a new service (e.g. as proxy endpoint), just invent a service ID, add it to the config and it will automatically spring into life wherever the config is read.

If the config mentions the same service ID twice in ways that don't make sense (e.g. two proxy endpoints have the same service ID), an error should be thrown.

ChristianSi:
This allows adding services to the config which will automatically spring into life if the config is read. Fixes #247.

### DIFF
--- a/thentos-core/src/Thentos.hs
+++ b/thentos-core/src/Thentos.hs
@@ -155,12 +155,12 @@ createDefaultUser st (Just (getDefaultUser -> (userData, roles))) = do
             else logger ERROR $ "failed to assign default user to roles: " ++ ppShow (UserId 0, result, user, roles)
 
 -- | Autocreate any services that are listed in the config but don't exist in the DB.
+-- Dies with an error if the default "proxy" service ID is repeated in the "proxies" section.
 autocreateMissingServices :: (db `Ex` DB) => ThentosConfig -> Action db ()
 autocreateMissingServices cfg = do
     dieOnDuplicates
     mapM_ (autocreateServiceIfMissing'P agent) allSids
   where
-    -- Die with an error if the default "proxy" service ID is repeated in the "proxies" section
     dieOnDuplicates  = case mDefaultProxySid of
         Just sid -> when (sid `elem` proxySids) . error $ show sid ++ " mentioned twice in config"
         Nothing  -> return ()
--- a/thentos-core/src/Thentos.hs
+++ b/thentos-core/src/Thentos.hs
@@ -32,13 +32,16 @@ import Data.Acid (AcidState, openLocalStateFrom, createCheckpoint, closeAcidStat
 import Data.Acid.Advanced (query', update')
 import Data.Configifier ((>>.), Tagged(Tagged))
 import Data.Either (isRight, isLeft)
+import Data.Maybe (maybeToList)
 import Data.Proxy (Proxy(Proxy))
 import System.Log.Logger (Priority(DEBUG, INFO, ERROR), removeAllHandlers)
 import Text.Show.Pretty (ppShow)
 
+import qualified Data.Map as Map
+
 import System.Log.Missing (logger, announceAction)
 import Thentos.Action
-import Thentos.Action.Core (ActionState(..), runActionWithPrivs, Ex)
+import Thentos.Action.Core (Action, ActionState(..), runActionWithPrivs, Ex)
 import Thentos.Config
 import Thentos.Frontend (runFrontend)
 import Thentos.Smtp (checkSendmail)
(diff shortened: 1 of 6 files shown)


ChristianSi:
DON'T MERGE YET!

It compiles but doesn't run.

When I do

```sh
$ cd thentos-adhocracy
$ cabal run thentos-adhocracy
```

I get:

```
... -- guardWrite:
    addServicePrim
    LabelError {lerrContext = ["guardWrite"], lerrFailure = "guardAllocP", lerrCurLabel = ("RoleAdmin" \/ "ServiceA (ServiceId {fromServiceId = \"qlX4MP7xEgtRng+8iNvMIcSo\"})") %% False, lerrCurClearance = "RoleAdmin" %% "RoleAdmin", lerrPrivs = [], lerrLabels = [("RoleAdmin" \/ "UserA (UserId {fromUserId = 0})") %% "RoleAdmin" /\ "UserA (UserId {fromUserId = 0})"]}
```

I don't know what's wrong, as the RoleAdmin role is set.

jkarni:
Tests maybe (esp. if there is in fact a run-time issue currently)?

fisx:
```
lerrCurLabel = ("RoleAdmin" \/ "ServiceA ... %% False
lerrCurClearance = "RoleAdmin" %% "RoleAdmin"
lerrLabels = [("RoleAdmin" \/ "UserA ...0) %% "RoleAdmin" /\ "UserA ...0"]
```

labels are what is passed to `guardWriteMsg`.  according to http://hackage.haskell.org/package/lio-0.11.5.0/docs/LIO-Core.html#g:6, `curLabel` must flow to `labels`, `labels` must flow to `curLabel`, and both must flow to `curClearance`.

It's simple enough to see why this fails.  But what should we change for it to not fail?

ChristianSi:
@jkarni  I'll add a test for the `autocreateServiceIfMissing'P` action.

ChristianSi:
> `curLabel` must flow to `labels`, `labels` must flow to `curLabel`

Hmm, to me it sounds as if OR (`\/`) checks are completely useless in that case since they'll effectively be turned into AND (`/\`) checks?

fisx:
Couldn't find an easy way to fix this.  we probably want to use `guardAllocMsg`, but we will also need to change some of the labels.

Given that the lio policy seems to be a bit brittle in general, I suggest to remove all guard*s and taints that cause trouble.  Then, independently of this PR, somebody should start writing tests that make the policy appearent by examples.  Those tests can be used to guide construction of a better policy (also in a different PR).

ChristianSi:
I outcommented the failing LIO check and now it works, unsurprisingly.

I still get errors when trying to log in A3, but that's because A3 tried to validate the X-User-Token and it's not there. Almost time to fix A3!

But first I'll add a test to complete this PR.

ChristianSi:
I added tests, it's now ready for review and (hopefully) merge.

thentos-core/src/Thentos.hs
```diff
@@ -149,3 +153,18 @@ createDefaultUser st (Just (getDefaultUser -> (userData, roles))) = do
...
             then logger DEBUG $ "[ok]"
             else logger ERROR $ "failed to assign default user to roles: " ++ ppShow (UserId 0, result, user, roles)
+
+-- | Autocreate any services that are listed in the config but don't exist in the DB.
+autocreateMissingServices :: (db `Ex` DB) => ThentosConfig -> Action db ()
+autocreateMissingServices cfg = do
+    dieOnDuplicates
+    mapM_ (autocreateServiceIfMissing'P agent) allSids
+  where
+    -- Die with an error if the default "proxy" service ID is repeated in the "proxies" section
```
jkarni: Could you move this comment to the haddocks? It's fine to throw an error on startup when first reading the config file, but there's always the chance that this function will be used elsewhere, where it might not be okay.

ChristianSi: I can to that, but it certainly doesn't make sense to use it, it's just a helper for the `makeMain` method. It's also not exported so it *cannot* be used elsewhere.

ChristianSi: I moved them comment.

ChristianSi:
Evidently Travis picked a version of hspec that's too old :( I'll check when the used functions were added...

Issue #212, opened 2015-07-14
Finish implementation of github single-sign-on for adhocracy.


fisx:
Some of this already happens in `Thentos.Adhocracy3.Backend.Api.Sso`, but it needs to be finished and tested.

Note that some Sso errors are still living in thentos-core, despite the fact that the Sso module is in thentos-adhocracy.  This should probably be cleaned up on the way.

fisx:
We will need to start changing the frontend code.  We should keep as much of the code as possible in a js library that the a3 frontend can load directly from thentos.  I see two options for us:

1. write something in purescript by hand.
2. generate javascript functions with servant-client.  we already have the api, so this may be the easier way, but it may be necessary to do things in addition to just rest calls.  perhaps we should generate purescript code with servant and write the missing bits in purescript.

fhartwig:
Finishing this would require significant changes to our current DB system. Since we're planning to move from acid-state to an RDBMS (see #268), I think this should be punted to avoid doing the work and then immediately throwing it away.

fisx:
We also need frontend hooks for this.  This is my current mental model (probably still has bugs):

```javascript
var ssoLoginButton = function(domElem, returnPath) { ... };
```

This takes a DOM element as argument and renders the button into that element.  It also installs an on-click handler that leaves the single-page app and does the sso handshake.  The `returnPath` argument is the url at which the session returns from the sso server (something like `/logged_in?state=...`).

```javascript
var ssoReturnPathHandler = function(returnPath, handler) { ... };
```

After the app is re-entered, there will be a way to update the app state based on the URL (the server doesn't know about app state).  This function lets the app developer register a function under the return path that takes no arguments and updates the state to "logged in".

I'm uncertain whether `returnPath` should be exposed to the user of this api, or whether it can be picked by thentos alone, without involvement of the app.  Or whether there are other credentials that need to be exposed.

I also suggest that we implement these functions in purescript.  It may be a bit of an overhead now, but we will have to write more frontend code soon.

We should discuss all this soon with somebody from a3 frontend!  @xi?

fisx:
I know I had done this before.  The `Sso` module header comment contains a more mature version of my last comment.

Btw @fhartwig, will that module also be moved to `thentos-core`?

fisx:
branch with more work: `2015-09-sso-experiment`

fisx:
closing this for now, since we finally have a requirements priority queue!

links:
- #315
[...]

PR #264, opened 2015-07-28
Move sso-related db fields, transactions and actions into thentos-adhocracy

fhartwig:
This is basically a refactoring for #212 , but since it's nice and self-contained I'd like to merge it before the actual implementation work.


jkarni:
Just so I understand - why move it to `thentos-adhocracy`?

fhartwig:
@jkarni Because that's the only place we actually want to use sso. The SSO module lives in that package, the transactions & actions were just in core for historical reasons.

jkarni:
Ok, I'm going to wait on travis, but otherwise LGTM

fhartwig:
@jkarni Sorry, forgot to add the new file. Fixing now...

PR #267, opened 2015-07-29
Use pre-built hlint instead of building it ourselves

### LINKED ISSUE #242: Don't build hlint on travis
Travis currently takes a long time to run. One of the problems is that we build hlint (including its dependencies) on every build. Since hlint depends on haskell-src-exts, which takes forever to build, this slows our build quite a bit.
I'm personally not convinced that we need to run hlint on travis, but even if we do, we should make sure it doesn't make our build take longer than necessary.

fhartwig:
Fixes #242 (don't merge yet, I want to see travis pass first)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,12 +15,15 @@ before_install:
     - make start
     - cd ../..
 install:
-    - cabal install hlint
     - ./misc/get-cabal-config.sh stack.yaml cabal.config
     - cp cabal.config thentos-core/cabal.config
     - cp cabal.config thentos-tests/cabal.config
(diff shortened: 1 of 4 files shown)


fisx:
It seems like your executable is not entirely self-contained...

```
hlint: /home/travis/hlint-build/.cabal-sandbox/share/x86_64-linux-ghc-7.8.4/hlint-1.9.21/Default.hs: openFile: does not exist (No such file or directory)

make[1]: *** [hlint] Error 123

make[1]: Leaving directory `/home/travis/build/liqd/thentos/thentos-core'

make: *** [thentos-core.hlint] Error 2

The command "make hlint" exited with 2.
```

PR #270, opened 2015-08-03
Build thentos-core with Werror too.

jkarni:


.travis.yml
```diff
@@ -26,5 +26,6 @@ install:
     - cp -r travis-cache/data ./hlint/.cabal-sandbox/share/x86_64-linux-ghc-7.8.4/hlint-1.9.21/
 script:
     - make hlint
+    - cd thentos-core && cabal configure --ghc-options="-Werror" && cabal build && cd ..
```
ChristianSi: Hm, if cabal fails then the final `cd ..` won't be executed and so the subsequent lines will fail too.

PR #272, opened 2015-08-03
A3 fails in a confusing way if the user path doesn't end with a slash

ChristianSi:
See https://github.com/liqd/adhocracy3.mercator/issues/1508


ChristianSi:
@robx I hope the importance will go away at some point.

PR #273, opened 2015-08-04
2015 07 jkarni test cleanup

jkarni:
Just leaving this PR here in case anyone wants to know where this is going.


jkarni:
OK, this still needs a rebase, and there are lots of other things I want to do with the tests, but it's probably best to consider merging already to avoid more conflicts and scope creep. The `notes.txt` file I'll remove before merging, but has some of the reasoning for things I did that was worth mentioning. Generally, I tried to start implementing pending tests that I could make sense of, and removed pending tests that looked outdated or whose intention I could no longer understand.

jkarni:
OK, done at last

robx:
There's some valuable comments in `notes.txt`, but they're sort of lacking context in that top-level file.

jkarni:
@robx I intend to remove them before merging (see comment above) - just notes as justification for things I did that might be controversial. If there's anything that's unclear in that context, let me know

robx:
@jkarni Ah, yes, missed that comment, thought this was meant to be merge-ready.

I don't have any particular issues with the potentially controversial notes/changes. The DB extension thing will go away when we switch to SQL, by the way.

Looks like a nice improvement to me overall.

jkarni:
I'm kind of annoyed at the hlint suggestion on travis (I can't see how we'd reduce duplication in a way that makes the code nicer). @robx @ChristianSi Should I just make a local binding anyhow, or remove that warning from HLint?

ChristianSi:
@jkarni I agree that eliminating such short pieces of duplication in test code (as opposed to production code) hardly makes sense. Currently `thentos-tests/HLint.hs ` contains the line

    ignore "Reduce duplication" = Thentos.FrontendSpec

You could change it to

    ignore "Reduce duplication"

to suppress that warning everywhere.

PR #275, opened 2015-08-04
Remove LIO restrictions to make helloworld service work. Fixes #231.

### LINKED ISSUE #231: 404 logging out from helloworld
logged in successfully as god/god, the clicked logout:

```
A web handler threw an exception. Details:
*** StatusCodeException
  Status { statusCode = 404 , statusMessage = "Not Found" }
  [ ( "Transfer-Encoding" , "chunked" )
  , ( "Date" , "Fri, 17 Jul 2015 15:48:39 GMT" )
  , ( "Server" , "Warp/3.0.13.1" )
  , ( "X-Response-Body-Start" , "thentos session not found" )
  , ( "X-Request-URL"
    , "DELETE http://localhost:7001/service_session/"
    )
  ]
  CJ { expose = [] }
```

robx:
This goes pretty far in removing LIO, this can't really be the right thing to do. Very unsure this should be merged.

It does however make the basic helloworld flow work:

* log in as god/god, create service, copy service id and token to helloworld's devel.config
* navigate to helloworld service, log in via thentos, redirected back
* log out of helloworld service

The last two steps also work as non-god. If we do decide to go ahead with this, I'd write tests for this before merging.

### DIFF
--- a/thentos-core/src/Thentos/Action.hs
+++ b/thentos-core/src/Thentos/Action.hs
@@ -360,9 +360,7 @@ allServiceIds = do
     query'P T.AllServiceIds
 
 lookupService :: (db `Ex` DB) => ServiceId -> Action db (ServiceId, Service)
-lookupService sid = do
-    taintMsg "lookupService" (RoleAdmin \/ ServiceA sid %% False)
-    query'P $ T.LookupService sid
+lookupService sid = query'P $ T.LookupService sid
 
 addService :: (db `Ex` DB) =>
     Agent -> ServiceName -> ServiceDescription -> Action db (ServiceId, ServiceKey)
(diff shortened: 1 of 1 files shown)


ChristianSi:
Looks OK for me but Travis has failed and refuses to show me why. Very annoying!

robx:
The change is sort of properly broken: It breaks the basic session-based authorization, which relies on lio. Sorry about that. It seems like we'd have to properly go over the lio stuff if we want both service and user sessions to work in some way...

ChristianSi:
Hmm, I'm still thoroughly confused about this whole LIO thing, wouldn't have expected an interaction of that kind...

fisx:
@ChristianSi: in 45fe690 above, travis chokes on compiler warnings.  in order to see them, you need to browse through the entire output, page by page.  (but perhaps there was another problem and this is not the answer to your confusion.  ignore me, i'm not here anyway :)

ChristianSi:
@fisx @robx When I said "Travis ... refuses to show me why" it was because I got a timeout when trying to view the Travis log. I've seen it now and there only seem to be two unused import warnings in `Thentos/Action.hs` that are trivial to fix.

If that fix is pushed I have no objections against merging this -- however, @robx says it breaks some other workflows though apparently there are no tests that can detect that?

robx:
@ChristianSi When running the tests locally I saw one scary test failure, which sounded like the change was circumventing an authorization check for basic user session functionality. Will have a more detailed look later.

jkarni:
@robx I think I came across the same issue. Is it being caused by the tests expecting Admin to not be able to lookup the session? I changed that in the test-cleanup PR since it didn't make much sense to me.

robx:
I've managed to restrict the LIO removals to service-related calls, now, so the change no longer affects thentos sessions.

PR #280, opened 2015-08-11
Better error responses

### LINKED ISSUE #274: Return proper error responses
Currently Thentos doesn't return proper HTTP replies in case of errors. There are two problems:

* No Content-Type header is set.
* The response body is just a naked error string.

E.g. to reproduce: start thentos-adhocracy and call

    curl -v -H 'X-User-Token: blah: blah' http://localhost:6546/

The response body will be "thentos session not found" without a Content-Type header.

The most simple, but hardly satisfying way would to be to add "Content-Type: text/plain" in case of errors. Then returning a naked error message would at at least result in a proper HTTP response.

But since the rest of the API is JSON, instead setting "Content-Type: application/json; charset=UTF-8" and wrapping the error message into a simple JSON object would probably be a better approach.
For example (somewhat similar to A3 error messages):

    {"status":"error", "error": "thentos session not found"}

Or (even simpler):

    {"error": "thentos session not found"}

In the case of the error situation mentioned here as example it's also confusing that the response code is HTTP/1.1 404 Not Found. It should rather be something like 403 Forbidden. But that's a detail for that specific error.

@fisx @jkarni @fhartwig @robx I'd like to hear your input on how to handle errors.

ChristianSi:
Fixes #274:

* Error responses now contain a JSON body with the Content-Type properly set.
* NoSuch... errors are reported as 400 Bad Request instead of 404 Not Found.

### DIFF
--- a/thentos-adhocracy/src/Thentos/Adhocracy3/Backend/Api/Simple.hs
+++ b/thentos-adhocracy/src/Thentos/Adhocracy3/Backend/Api/Simple.hs
@@ -359,7 +359,7 @@ addUser (A3UserWithPass user) = AC.logIfError'P $ do
     sendUserConfirmationMail (Tagged $ cfg >>. (Proxy :: Proxy '["smtp"])) user activationUrl
     let userPath = userIdToPath cfg uid
         typedPath          = TypedPath userPath CTUser
-        -- Mimic cache-control info returned form the A3 backend
+        -- Mimic cache-control info returned from the A3 backend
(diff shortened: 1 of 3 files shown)


ChristianSi:
I just noticed that there are thentos-adhocracy specific errors that need to be adapted as well. I'll add another commit to do that. Didn't know we had already modularized that...

jkarni:
LGTM. Waiting on travis

PR #281, opened 2015-08-14
Switch transactions to use postgresql-simple.

robx:
thentos-core now compiles, with a lot of gutted transactions and no database. But getting there.


jkarni:
I imagine most/all of these are going away, but you can use [file-location](https://hackage.haskell.org/package/file-location) for error locations that don't have to be kept up-to-date. (7.10 also has more explicit support.)

robx:
Two failing test cases left in thentos-tests, bad header test goes against a removed end-point, and creating a session for a non-existing user should fail with NoSuchUser but currently throws an SQL foreign key error.

jkarni:
> Is this automatically transactional? We shouldn't confirm the user and then throw an error.

@robx As in, does postgresql-simple make multiple SQL statements (in a single `Query`) a single transaction? If so, then yes.

robx:
@jkarni Exactly. Thanks, didn't really expect that.

jkarni:
What's left to do here?

robx:
The tests pass, so... merge? But:

* Personally, I'd like to get rid of the `FromRow` instances before merging.
* It would be a good idea to run the core executable (and maybe the helloworld service) and see what fails. Then write some tests for that stuff.
* Regarding not-done things: there's a bit of a mess with users/services vs. agents, with some things only implemented for users.
* Update the README (postgres dependency, in particular).
* I wouldn't like to merge this with stubbed transactions -- either we find a need to implement them, or we should remove them.

robx:
From my perspective, this is good enough to merge. I've checked that the thentos-core executable works (except for https://github.com/liqd/thentos/issues/300, but that's an LIO issue which might also affect master, and we need to work on that stuff anyway). The User/Service/Agent thing is still open, but again I think that's something we can improve in master.

PR #282, opened 2015-08-17
A3 style error handling

ChristianSi:
This makes adhocracy-thentos emit errors that like looks those from the A3 backend, so that the frontend will understand them. For all typical error situations that can occur with user creation, account activation, or login, the errors mimic those from the backend exactly.

This means that not only the happy path, but also the usual unhappy paths now work as they should without requiring any further changes at the side of A3.

### DIFF
--- a/thentos-adhocracy/src/Thentos/Adhocracy3/Types.hs
+++ b/thentos-adhocracy/src/Thentos/Adhocracy3/Types.hs
@@ -26,13 +26,16 @@ module Thentos.Adhocracy3.Types
 import Control.Applicative ((<$>))
 import Control.Exception (Exception)
 import Control.Lens (makeLenses)
+import Data.Aeson (Value(String), ToJSON(toJSON), (.=), encode, object)
 import Data.Data (Typeable)
+import Data.Maybe (fromMaybe)
 import Data.SafeCopy (SafeCopy, deriveSafeCopy, base, putCopy, getCopy)
 import Data.Set (Set)
 import Data.String.Conversions (LBS, ST)
 import Data.Thyme.Time () -- required for NominalDiffTime's num instance
 import GHC.Generics (Generic)
-import Servant.Server.Internal.ServantErr (err500)
+import Servant.Server (ServantErr)
+import Servant.Server.Internal.ServantErr (err400, err500, errBody, errHeaders)
 import System.Log.Logger (Priority(ERROR))
 
 import Thentos.Backend.Core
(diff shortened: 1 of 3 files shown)


thentos-core/src/Thentos/Backend/Core.hs
```diff
@@ -80,13 +81,15 @@ newtype ErrorMessage = ErrorMessage { fromErrorMessage :: ST }
 instance ToJSON ErrorMessage where
     toJSON (ErrorMessage msg) = object ["status" .= String "error", "error" .= msg]
 
+-- |Type of functions that transform a base error and an error message into a ServantErr.
+type MkServantErrFun = ServantErr -> ST -> ServantErr
```
jkarni: I prefer not using type synonyms here, since it's much less obvious what type `mkServantErr` actually expects by looking just at its signature.

thentos-adhocracy/tests/Thentos/Adhocracy3/Backend/Api/SimpleSpec.hs
```diff
@@ -91,6 +95,40 @@ spec =
...
+                        "\"User doesn't exist or password is wrong\""
+
+        describe "activate_account" . before setupTestBackend . after teardownTestBackend $
+            it "rejects bad path mimicking A3" $
+                \bts -> runTestBackend bts $ do
+
+                    let reqBody = encodePretty . object $
+                            [ "path" .= String "/activate/no-such-path" ]
+
+                    rsp <- srequest $ makeSRequest "POST" "activate_account" [] reqBody
```
jkarni: I don't entirely understand why we've been using `srequest` and `makeSRequest`. In this particular case, wouldn't something like

```
context "when the path doesn't exist" do
  it "responds with 400" $ do 
     post "activate_account" reqBody `shouldRespondWith` 400
  it "sets the content-type to application/json" $ do 
     post "activate_account" reqBody `shouldRespondWith` ( "Content-Type" ...)
  it "responds with the appropriate body" $ do
    post "activate_account" reqBody `shouldRespondWith` [json|...]
```
Test everything that this does (plus a little more about the actual json structure), be a little cleaner (in the sense that we'd have smaller orthogonal tests, and a more self-evident specification of the behaviour we expect), and also be shorter?

thentos-core/src/Thentos/Backend/Core.hs
```diff
@@ -114,52 +117,58 @@ actionErrorToServantErr e = do
 
 
 class ThentosErrorToServantErr db where
-    thentosErrorToServantErr :: ThentosError db -> (Maybe (Priority, String), ServantErr)
+    -- | Convert a 'ThentosError' into a 'ServantErr'. If the first argument is 'Just' provided,
+    -- it must be invoked to create the returned error, otherwise a suitable default implementation
+    -- must be used for that purpose. This allows derived db's to adapt the details of the error
+    -- actually returned.
+    thentosErrorToServantErr :: Maybe MkServantErrFun -> ThentosError db
```
jkarni: Do we really need the extra `Maybe` argument here? It seems like if a derived db wanted to change the error, it could already do it (by applying whatever function [here](https://github.com/liqd/thentos/pull/282/files#diff-9520a252d8a4df0b5abfefea674c70f7R150)).

thentos-core/src/Thentos/Backend/Core.hs
ChristianSi: I disagree. The type synonym is used not only in the type of `mkServantErr`, but also in the type of `thentosErrorToServantErr` and in `Thentos/Adhocracy3/Types.hs` where an A3-specific MkServantErrFun is defined. Using the type synonym instead of copy-and-pasting the underlying types multiples times makes the code more compact and self-documenting.

It's also advantageous should we ever have reason to change to type of that function -- just check where the type synonym is used to get a grasp of what might need to be changed.

thentos-adhocracy/tests/Thentos/Adhocracy3/Backend/Api/SimpleSpec.hs
ChristianSi: I'm not sure what you mean with the `[json|...]` in the last line but I would suppose that would result in a bit-for-bit comparison that the encoded JSON matches the response body? I feel that would be testing too much---the same JSON can be encoded in various ways, so we shouldn't expect that all fields are written in the same order etc.

Also I think that for each `it` we would have to do a separate `runTestBackend bts`, so not only would that actually be quite long-ish, but having too many separate `it`s would also slow things down.

I know that some people swear to the style "test just *one thing* per test" and I agree that that's reasonable in theory, but considering that our test suite is already quite slow I think we shouldn't be too pedantic about that in practice.

thentos-core/src/Thentos/Backend/Core.hs
ChristianSi: Yes it's needed, how else should the A3-specific implementation ensure that *all* errors have the A3-specific format? Calling this function without that argument would result in a ServantErr whose JSON body already contains the wrong kind of JSON data structure.

Of course, the derived db could then decode that JSON data structure and repackage the contained information using it's own custom style, but that would be more than ugly.

ChristianSi:
I would like to merge as is. @jkarni, can you agree?

thentos-core/src/Thentos/Backend/Core.hs
jkarni: > how else should the A3-specific implementation ensure that all errors have the A3-specific format

If this is what we're concerned with, I'd just opaquely newtype-wrap the error, and have a smart constructor (which would also fix the problem that right now `ServantErr`s can have their body and headers overwritten by `mkServantErr`, and nothing in the types indicates one should be worried about that). Then the types would guarantee the invariant we're concerned with is fulfilled. But in this case it seems to me like only the A3 DB type can be used in A3, so I'd really just apply the transformation once in the line I mentioned and be done with it. I still don't see how that's a weaker guarantee. 

More generally, if I have an `a :: Int`, and often apply a function `fn :: Int -> SomeType -> Int` to it, I wouldn't consider changing the type of `a` to `a :: (Int -> SomeType -> Int) -> Int`. It's more confusing and complicated, and there's nothing I can do with that that I couldn't do already.

thentos-adhocracy/tests/Thentos/Adhocracy3/Backend/Api/SimpleSpec.hs
jkarni: OK, good point about `[json|...]`. I still think we could use `shouldSatisfy` or whatever and actually test exactly what we want rather than just substring, since the specification we want is that it be a particular json format, *not* that it be valid json and contain certain substrings.

I'm trying to improve the speed of the test suite, and soon the overhead of setting up each test should be negligible (at least until we move to some other DB), and to moreover be able to run them in parallel. It is very hard currently to tell from our tests what is the specification we're trying to match against, and what just happened to be expedient (or, even more importantly, not expedient) to test. I really don't think we should be skimping on making our tests readable and complete.

thentos-core/src/Thentos/Backend/Core.hs
jkarni: OK, I agree this is mostly a personal preference issue :).

thentos-adhocracy/src/Thentos/Adhocracy3/Types.hs
```diff
@@ -82,20 +85,81 @@ instance SafeCopy (ThentosError DB)
...
+        -- exactly so that the frontend recognizes them
+        f (ThentosA3ErrorCore e) = case e of
+            BadCredentials -> (Nothing, mkA3StyleServantErr err400 $ A3ErrorInfo
+                "password"
+                "body"
+                "User doesn't exist or password is wrong")
+            UserEmailAlreadyExists -> (Nothing, mkA3StyleServantErr err400 $ A3ErrorInfo
+                "data.adhocracy_core.sheets.principal.IUserExtended.email"
+                "body"
+                "The user login email is not unique")
```
jkarni: None of this looks like it's being tested.

ChristianSi: I test the conversion of `BadCredentials` and `NoSuchPendingUserConfirmation`, not of all the A3-specific error types. If those two work, it's reasonable to assume that the others will work as well.

thentos-adhocracy/tests/Thentos/Adhocracy3/Backend/Api/SimpleSpec.hs
ChristianSi: Well I thought about importing the JSON class used to generate error and check that we can convert the serialized JSON back to it, but that felt a bit like cheating. It drilling into JSON *without* a defining type class is no fun at all (I did a bit of that in other tests).

So I choose to just make sure that it *is* JSON and contains the specifically relevant bits of information (including the surrounding quotation marks added by JSON). For me that's good enough, but if you want to improve the tests. feel free to do so.

Issue #239, opened 2015-07-21
Fix commented-out tests


fisx:
`thentos-adhocracy/tests/Thentos/Adhocracy3/Backend/Api/SimpleSpec.hs` contains a test that now requires an a3 backend to run.  Either find a way to run a3 backend for the tests or (much preferrred) mock the a3 backend somehow so that we can run the test without the real thing.

This depends on #238.

PR #286, opened 2015-08-19
Activate users in A3 and Thentos at the same time.

### LINKED ISSUE #283: Thentos proxy: handle user activation properly
Right now user activation in A3 is turned off if it runs behind a Thentos proxy. In A3, all created users are immediately considered active. Thentos sends the activation email and activates the account after the user has clicked on a link.

This is not a good solution since only active users are visible and listed, and if somebody tries to access a not-yet-activated user, they get an error response (410 Gone). By activating all users immediately (in A3), the Thentos proxy bypasses these privacy settings.

To fix this, we should activate users in A3 at the same time as in Thentos. The most simple solution to do that seems to be:

* Let A3 send activation emails rather than do it in Thentos. This also has the advantage that A3 defines templates that make it easy to change the subject and contents of these emails, while in Thentos they are currently hardcoded.
* If the user clicks on an activation link, rather then handling the activation completely in Thentos, forward the request to A3 which will respond with 200 or an error. If 200, activate the user in Thentos too. (The response sent by A3 reveals which user was just activate). Then forward the response to the client.

ChristianSi:
Fixes #283.


thentos-core/src/Thentos/Transaction/Transactions.hs
```diff
@@ -203,6 +204,20 @@ trans_finishUserRegistration now expiry token = polyUpdate $ do
...
     return uid
 
+-- | Confirm a user based on the 'UserId' rather than the 'ConfirmationToken.'
+trans_finishUserRegistrationById :: (db `Extends` DB) =>
+    Timestamp -> Timeout -> UserId -> ThentosUpdate db ()
+trans_finishUserRegistrationById now expiry uid = polyUpdate $ do
+    matchingUnconfirmedUsers <- Map.filter unconfirmedUserMatches <$> gets (^. dbUnconfirmedUsers)
+    case Map.keys matchingUnconfirmedUsers of
+        [tok] -> void $ trans_finishUserRegistration now expiry tok
+        []    -> throwT NoSuchUser
```
jkarni: #284 switched this to `throwError`, which is why travis isn't passing

ChristianSi: OK, I pushed a fix, we'll see what happens now...

thentos-adhocracy/tests/Thentos/Adhocracy3/Backend/Api/SimpleSpec.hs
```diff
@@ -108,16 +108,26 @@ spec =
...
-
-                    rsp <- srequest $ makeSRequest "POST" "activate_account" [] reqBody
+                    let headers = [("X-User-Token", "invalid-token")]
+                    rsp <- srequest $ makeSRequest "POST" "dummy/endpoint" headers ""
                     shouldBeErr400WithCustomMessage (simpleStatus rsp) (simpleBody rsp)
-                        "\"Unknown or expired activation path\""
+                        "\"Invalid user token\""
+
+-- Disabled because account activation now requires the A3 backend to run
+--        describe "activate_account" . before setupTestBackend . after teardownTestBackend $
```
jkarni: Starting a destination server that always return a `RequestResult` (`\_ respond -> respondLBS $ encode myReqResult`) should suffice to make this test work, no?

thentos-core/src/Thentos/Transaction/Transactions.hs
```diff
@@ -21,6 +21,7 @@ import Control.Monad.State (modify, gets, get)
 import Data.AffineSpace ((.+^))
 import Data.EitherR (catchT, throwT)
 import Data.Set (Set)
+import Data.String.Conversions ((<>))
```
jkarni: I find it confusing to import this from here rather than `Data.Monoid`.

thentos-adhocracy/src/Thentos/Adhocracy3/Backend/Api/Simple.hs
```diff
@@ -452,6 +486,9 @@ extractUserId resp = do
         (Aeson.eitherDecode . Client.responseBody $ resp :: Either String TypedPath)
     userIdFromPath $ tpPath resource
 
+sendRequest :: Client.Request -> IO (Client.Response LBS)
+sendRequest req = Client.withManager Client.defaultManagerSettings $ Client.httpLbs req
```
jkarni: The documentation for the `http-client` library says using a new manager for each request to the same host isn't great (also `withManager` is being deprecated). We already have a manager being threaded around - can't we use that one?

thentos-adhocracy/src/Thentos/Adhocracy3/Backend/Api/Simple.hs
```diff
@@ -429,21 +452,32 @@ a3corsPolicy = CorsPolicy
...
     defaultProxy <- Tagged <$> config >>. (Proxy :: Proxy '["proxy"])
     let target = extractTargetUrl defaultProxy
-        user'  = UserFormData { udName     = udName user,
-                                udEmail    = udEmail user,
-                                udPassword = "dummypass" }
-    initReq <- Client.parseUrl $ cs $ show target <> "/principals/users"
-    return initReq { Client.method = "POST",
-        Client.requestHeaders = [("Content-Type", "application/json")],
-        Client.requestBody = Client.RequestBodyLBS . Aeson.encode . A3UserWithPass $ user' }
+    initReq <- Client.parseUrl $ cs $ show target <> route
```
jkarni: Is there a way of verifying the url works at start-up (or even compile) time? Having a lot of partial functions at other points seems like bad practice (even if we'd notice the error early during runtime).

thentos-adhocracy/tests/Thentos/Adhocracy3/Backend/Api/SimpleSpec.hs
ChristianSi: Not sure what you mean by `myReqResult` but if we have a fake A3 backend that returns a suitable response that would do the trick, yes. That's too big for this PR, though, there already a separate ticket for it (#239). If you have a good idea for how to do that, by all means start working on that ticket!

thentos-adhocracy/src/Thentos/Adhocracy3/Types.hs
```diff
@@ -127,28 +147,30 @@ instance ThentosErrorToServantErr DB where
...
                 "Unknown or expired activation path")
-            NoSuchThentosSession -> (Nothing, mkA3StyleServantErr err400 $ A3ErrorInfo
+            NoSuchThentosSession -> (Nothing, mkA3StyleServantErr err400 $ A3Error
                 "X-User-Token"
                 "header"
                 "Invalid user token")
             _ -> thentosErrorToServantErr (Just mkErr) e
 
+        f (GenericA3Error errMsg) =
+            (Nothing, err400 {errBody = encode errMsg, errHeaders = [contentTypeJsonHeader]} )
```
jkarni: It seems like `400` can be misleading here. E.g. if A3 is down we probably want 502. I know you don't want to fully parse the possible errors, but would relaying the status code from A3 suffice?

thentos-adhocracy/tests/Thentos/Adhocracy3/Backend/Api/SimpleSpec.hs
jkarni: `myReqResult = RequestSuccess x y` (not really sure what `x` and `y` would be). We already set up a destination server in https://github.com/liqd/thentos/blob/master/thentos-adhocracy/tests/Thentos/Adhocracy3/Backend/Api/ProxySpec.hs ; this one would be quite a bit simpler, though.

Tests get outdated when commented out. #239 is going to be a lot of work for that reason, not because setting up a destination server is that hard.

thentos-core/src/Thentos/Transaction/Transactions.hs
ChristianSi: We often do that. I fixed it in that specific case, through.

thentos-adhocracy/src/Thentos/Adhocracy3/Backend/Api/Simple.hs
ChristianSi: I don't know, I think you made the refactoring to use a single manger so you should know how to do that. Maybe you can change that code in a separate PR? It wouldn't belong in this PR anyway, as that code is old(er) -- I just moved it around a bit.

thentos-adhocracy/src/Thentos/Adhocracy3/Backend/Api/Simple.hs
ChristianSi: Functions that invoke an external service are "partial" by default. We could make sure that the service is reachable at startup, but it could always go down later or the network connection could break down. So that can't be helped.

jkarni: Sorry, I was talking about the `parseUrl` line (and the `defaultProxy` line too, I guess), not the request itself.

jkarni: (The answer may be that there's no neat way of doing this - I just think we should try.)

jkarni: Unrelatedly, `<>` seems like a dangerous way of concatenating paths/urls. It could result in multiple or missing slashes.

thentos-adhocracy/src/Thentos/Adhocracy3/Types.hs
ChristianSi: When A3 is down (= not reachable or doesn't respond) we'll throw a 500 instead of this GenericA3Error anyway. To requests such as this one A3 *always* replies with 400 in case of problems, so 400 should be safe here.

In rare cases (= internal problem) A3 could respond with 500 instead of 400, but those cases will still be distinguishable by the details of the A3 error message which we pass along. Along, the A3 frontend doesn't care for such details anyway in the case of activation clicks. It just displays the same "sorry that went wrong" method in case of *any* errors without caring for such details as status code or error message.

So it won't make a difference here.

jkarni: OK

thentos-adhocracy/src/Thentos/Adhocracy3/Backend/Api/Simple.hs
```diff
@@ -403,17 +417,26 @@ login r = AC.logIfError'P $ do
...
   where
-    sendRequest ::  Client.Request -> IO (Client.Response LBS)
-    sendRequest req = Client.withManager Client.defaultManagerSettings $ Client.httpLbs req
     responseCode = Status.statusCode . Client.responseStatus
 
+-- | Activate a user in A3 and return the response received from A3.
+activateUserInA3'P :: ActivationRequest -> AC.Action DB RequestResult
+activateUserInA3'P actReq = do
+    config <- AC.getConfig'P
+    let a3req = fromMaybe (error "activateUserInA3'P: mkRequestForA3 failed, check config!") $
```
jkarni: If I understand correctly, this error could happen either because (1) the config file is wrong, (2) A3 is down. I think the correct http status code in the latter case is 502, not 500.

ChristianSi: No that error can only occur if the config is wrong, therefore it's properly an `error` (which will always result in status 500).

jkarni: I see. Fair enough!

thentos-adhocracy/src/Thentos/Adhocracy3/Backend/Api/Simple.hs
ChristianSi: Yeah, it's a bit lax and could lead to `//` instead of `/`. In practice that shoudn't hurt, though.

ChristianSi: As for `parseUrl`, I'm not really sure how we can change that. All we can access as global state are the config and the DB. The parsed URL certainly doesn't belong in the DB, and in the config there are currently only fields which are really explicitly stored there. Might be worthwhile considering changing that in some way but I think that needs some more pondering.

jkarni: > Yeah, it's a bit lax and could lead to // instead of /. In practice that shoudn't hurt, though.

:)

I'm still in favour of defining `a </> b = (reverse $ dropWhile (== '/') $ reverse a) <> "/" <> dropWhile (== '/') b` (or using it from some package).

ChristianSi: I will look into that in a separate PR.

fhartwig: We have the `(<//>)` operator defined in `thentos-core/src/Thentos/Util.hs`, which seems to be what you want.

jkarni: And apparently we also use `</>` from `System.FilePath`

PR #287, opened 2015-08-25
generalized error type

robx:
This puts some flexibility back into the errors, which got lost when DB extension was removed. Also tries to clean up the conversion to servant errors a bit. Something like this will be needed to get thentos-adhocracy to work with postgres.

I'm not entirely convinced this is the best way to go. Problems:
* `ThentosError ()` now has a bogus error `OtherError ()`
* need `(Show e, Typeable e)` contexts all over the place
* need to add type signatures to specialize actions to `Action () a` from `Action e a`, particularly in tests; compare the the runAction mess in ActionSpec.hs

The alternative approach that I see is to have

```
-- no ThentosError in here
newtype Action e a = Action (ReaderT ActionState (EitherT e (LIO DCLabel)) a)

type ThentosAction = Action ThentosError
```
then define all transactions/actions in thentos-core as ThentosAction and friends.

In thentos-adhocracy, define `ThentosA3Error` as before to extend `ThentosError`, and "lift" actions via
```
Action ThentosError a -> Action ThentosA3Error a
```

### DIFF
--- a/thentos-core/src/Thentos.hs
+++ b/thentos-core/src/Thentos.hs
@@ -55,7 +55,7 @@ import Thentos.Transaction.Core (runThentosQuery)
 -- * main
 
 main :: IO ()
-main = makeMain $ \ (actionState@(ActionState (st, _, _))) mBeConfig mFeConfig cmd ->
+main = makeMain $ \ actionState mBeConfig mFeConfig cmd ->
     case cmd of
         Run -> do
             let backend = maybe (return ())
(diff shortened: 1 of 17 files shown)


jkarni:
>  ThentosError () now has a bogus error OtherError () 

Use `Void` instead of `()`?

But your suggestion also sounds good. Is the idea to still have `type ThentosError = ThentosError' Void` and `type ThentosA3Error = ThentosError' A3Error`, or have entirely different types?

robx:
@jkarni I was thinking `ThentosError` with no parameter, and `ThentosA3Error` with a constructor `CoreError ThentosError` as in master. But let's just go with this for now to get things to work.

PR #292, opened 2015-09-07
Handle password reset in Thentos proxy

### LINKED ISSUE #168: A3 API: Handle User Password reset
The A3 backend implements an endpoint that allows resetting the passwords of users, see the last section of https://github.com/liqd/adhocracy3.mercator/blob/master/src/adhocracy_core/docs/authentication_api.rst. Our A3 compatibility API doesn't yet. We won't need that for SSO, but sooner or later we'll need it for full compatibility with the A3 frontend.

ChristianSi:
Fixes #168.


fhartwig:
:+1:

Issue #285, opened 2015-08-18
Endpoints fall through


jkarni:
When a path matches but the request body or content-type doesn't, the next endpoints get tried. This goes against the idea of preventing all requests under certain paths from going to the proxied service.

ChristianSi:
My understanding was that that's a bug in Servant?

jkarni:
The content-type stuff is, and I opened a ticket for it there. The request-body is less clear (I would say no), and we rely on the current behaviour in thentos.

ChristianSi:
I'd say that if the request body is not parsable, Servant should respond with an error. At least, that's how it should be in our use case.

ChristianSi:
This really is an extremely serious problem and needs to be fixed before we can even think of going live anywhere. Right now I work on A3 password reset (#168). As usual in proxy mode, we handle some endpoints ourselves and forward the request to A3. But if we cannot successfully parse a request body intended for one of our own endpoints (e.g. our FromJSON instance rejects a password because it's too short), then that endpoint is discarded and Servant instead calls the "catch-all" serviceProxy endpoint, thus forwarding the request to A3.

In that way, the A3 backend will see data it was never intended to see!

We cannot run the Thentos proxy in such a broken setup. Rather, all errors, no matter where they are thrown, **must** abort request processing and be reported to the user in the appropriate way.

Not only is this a show-stopper that makes actual deployment impossible, it also makes realistic testing of error situations impossible because the error will simply be discarded. This error must therefore be fixed with the utmost priority.

fisx:
@ChristianSi: agreed, thanks for pinging this.  we'll look into it.

@jkarni: you said you opened a ticket "there".  do you mean haskell-servant/servant#41?  do you have time for this later today?

fisx:
See aso: haskell-servant/servant#214

@ChristianSi, could you take a look at make sure the test cases model the behavior we need?  Did I miss anything?

ChristianSi:
@fisx I replied there.

jkarni:
Bad request done in #294.

jkarni:
The overall picture is pretty [grim](https://github.com/haskell-servant/servant/pull/218). I don't think fixing all of this will be too hard, though.

PR #294, opened 2015-09-07
Use fixed 400 error servant branch

jkarni:
See https://github.com/haskell-servant/servant/compare/jkarni/400-error-fix?expand=1 for the actual changes.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -12,8 +12,6 @@ packages:
   extra-dep: true
 - location: 'submodules/servant/servant-docs'
   extra-dep: true
-- location: 'submodules/servant/servant-jquery'
-  extra-dep: true
 extra-deps:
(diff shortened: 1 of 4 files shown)


jkarni:
See #285

fisx:
:+1:

thanks (-:

ChristianSi:
Travis is unhappy:

```
src/Thentos/Backend/Core.hs:46:33:
    Module ‘Servant.Server.Internal’ does not export ‘methodRouter’
src/Thentos/Backend/Core.hs:46:47:
    Module ‘Servant.Server.Internal’ does not export ‘Router(..)’
```

IIRC, these functions where added in an as-yet-unreleased commit to Servant master. Maybe the branch is derived from the latest Servant release rather than from master?

PR #297, opened 2015-09-08
stop fake a3 backend even on test failure

robx:
I'm not really sure this is the best way to do things, but it's the only way I could figure out.

Somehow, it's wrong to take the a3 message out of the test `it`, but `around` and friends only work on that level. An alternative might be to run `bracket` inside the `it`, but that's tricky with the `WaiSession`.


ChristianSi:
Looks good, will merge if Travis is happy.

ChristianSi:
Merged into #296, therefore this can be closed.

Issue #261, opened 2015-07-24
"stack bench" fails


robx:
```
$ stack bench
While constructing the BuildPlan the following exceptions were encountered:

--  While attempting to add dependency,
    Could not find package pronk in known packages

--  Failure when adding dependencies:    
      pronk: needed (>=0.1.1), but not present in build plan
    needed for package: thentos-tests-0.0.1
```

This is how far I got, which gave me pages of dependency errors:

```
diff --git a/stack.yaml b/stack.yaml
index 029c1b1..770982d 100644
--- a/stack.yaml
+++ b/stack.yaml
@@ -4,6 +4,8 @@ packages:
 - 'thentos-tests'
 - 'thentos-adhocracy'
 - 'services/helloworld'
+- location: 'submodules/pronk'
+  extra-dep: true
 - location: 'submodules/servant/servant'
   extra-dep: true
 - location: 'submodules/servant/servant-server'
@@ -17,13 +19,16 @@ packages:
 extra-deps:
 - acid-state-0.12.4
 - configifier-0.0.5
+- criterion-0.6.2.1
 - cryptonite-0.5
 - digestive-functors-0.8.0.0
 - digestive-functors-blaze-0.6.0.6
 - digestive-functors-snap-0.6.1.3
 - email-validate-2.1.3
 - functor-infix-0.0.3
+- hastache-0.5.1
 - lio-0.11.5.0
+- meldable-heap-2.0.3
 - regex-easy-0.1.0.0
 - regex-pcre-0.94.4
 - sandi-0.3.5
@@ -31,6 +36,8 @@ extra-deps:
 - snap-blaze-0.2.1.3
 - snap-loader-static-0.9.0.2
 - snaplet-acid-state-0.2.7
+- statistics-0.10.5.2
+- text-1.2.1.1
 - uri-bytestring-0.1.6
 
 resolver: lts-2.18
```

jkarni:
Doing the same thing from `pronk` that we do for `servant` should fix it.

fisx:
@robx: i think this is fixed in #301.  if not, please re-open.

PR #301, opened 2015-09-14
use ghc-7.10; add version constraints to cabal files; fix benchmarking

fisx:
- I used https://github.com/fisx/cabal-extras for generating constraints based on stackage and the old freeze files.
- Then I ended up making a lot of changes to the code to kill 7.10 warnings and errors (and some from newer-than-expected libs).
- Updated stack to get it to work with the new package files.
- pronk needed some work to cooperate with criterion >1.  `cabal bench` is now spitting out (dubious yet consistent) figures again.


jkarni:
What's the point of such strict version bounds? It doesn't seem very different from just having a cabal freeze file?

fisx:
freeze files:
- can't be relaxed as you go
- add an extra hack onto the pile, especially in deployment

is that enough reason?  why/where does the tightness bother you?

jkarni:
> is that enough reason?
Yup :)

fisx:
This should also take care of #261.

.travis.yml
```diff
@@ -1,10 +1,11 @@
 language: haskell
-ghc: 7.8
 before_install:
+    - grep '\(MemTotal\|SwapTotal\)' /proc/meminfo
     - travis_retry sudo add-apt-repository -y ppa:hvr/ghc
     - travis_retry sudo apt-get update
-    - travis_retry sudo apt-get install --force-yes happy-1.19.3 alex-3.1.3 cabal-install-1.22
-    - export PATH=/opt/alex/3.1.3/bin:/opt/happy/1.19.3/bin:/opt/cabal/1.22/bin:$PATH
+    - travis_retry sudo apt-get install --force-yes ghc-7.10.2 happy-1.19.3 alex-3.1.3 cabal-install-1.22
+    - cat /var/lib/dpkg/info/ghc-7.10.2.list
```
jkarni: What's the point of this?

README.md
```diff
@@ -117,6 +117,17 @@ required if you want to run tests. Whether you need other packages will
 depend on your use case.
 
 To run executables or tests, you will need to install PostgreSQL.
+Depending on your setup, you may need to cast some authorization
+spells.  Here is what works on debian (run as user postgres):
+
+```shell
+[as user postgres] $ createuser thentos -d
```
jkarni: to be fully explicit, maybe remote the `as user` stuff and say instead:

`sudo -u postgres createuser ...`

thentos-core/src/Thentos/Transaction.hs
```diff
@@ -1,11 +1,11 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE FlexibleContexts #-}
```
jkarni: Is this right? How was it compiling before?

thentos-core/thentos-core.cabal
```diff
@@ -43,7 +43,7 @@ library
   hs-source-dirs:
       src
   ghc-options:
-      -Wall
+      -Wall -j5
```
jkarni: I think this should be left for the user to decide at the `cabal configure` stage.

thentos-tests/bench/Main.hs
```diff
@@ -36,35 +41,31 @@ import Thentos.Types
...
 
 
 main :: IO ()
 main = do
-    fts <- setupTestServerFull
-    let cfg = fts ^. ftsCfg
-    putStrLn $ ppShow cfg
-    threadDelay $ let s = (* (1000 * 1000)) in s 2
+    withFrontendAndBackend "thentos_bench" $ \ as@(ActionState (_, _, cfg)) -> do
+        -- putStrLn . cs . renderConfigFile $ cfg
```
jkarni: remove?

thentos-core/thentos-core.cabal
fisx: Ok, but isn't 5 a more sensible default than 1?  (I'm assuming you can override this with `configure`?)

thentos-tests/bench/Main.hs
fisx: ok.

thentos-tests/bench/Main.hs
```diff
@@ -1,26 +1,29 @@
-{-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE ViewPatterns      #-}
+{-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE ViewPatterns        #-}
+{-# LANGUAGE ScopedTypeVariables #-}
```
jkarni: Is this being used?

thentos-tests/src/Thentos/Test/Config.hs
```diff
@@ -27,10 +27,14 @@ command: "run"
 backend:
     bind_port: 7118
     bind_host: "127.0.0.1"
+    expose_port: 7118
```
jkarni: Can it be documented somewhere what this does?

thentos-tests/bench/Main.hs
fisx: yes.

fisx:
back to you @jkarni. (is this good documentation?)

jkarni:
Looks good to me. Waiting on travis.

fisx:
@jkarni: I didn't know you would look into this today, so I've bothering @robx.  Before you merge: we are working on a travis issue.

In some runs, cabal is killed when building dependencies (most of the time it is SafeCopy, which, now that I think of it, should not be in the list any more anyway).  Setting the heap size to 500M in `./misc/thentos-install.sh` doesn't help, either because it has no effect, or because the problem lies elsewhere.  I have set the number of cores to use for parallel module compilation back to `1` (the default), and will run travis a few more times.  I will add another comment here with the results.

fisx:
Hm, still crashing.  `-j1` does not appear to have anything to do with it.  :-(

I guess the next step would be to turn off parallelism on the package level.  Since I think we only have 1, sometimes 2 cores, that may not even be too bad for travis roundtrip times.  I'll try that later and report back here.

Possibly relevant:

- https://ghc.haskell.org/trac/ghc/ticket/10228
- http://permalink.gmane.org/gmane.comp.lang.haskell.glasgow.user/25533
- https://www.reddit.com/r/haskell/comments/33eyi3/ghc_710_and_memory_usage/

There is talk about RAM issues with 7.10 and machines with 2GB or 4GB.

fisx:
I'm not happy with travis.

1. when pushing a new commit while the previous one is still waiting for a host to run on, that previous one will illegally pick the new commit for running.  very confusing.
1. the commit hashes in the travis logs are wrong, even in cases where appearently the right commit was picked.
1. sometimes, travis won't notice when the tests successfully complete, and hang.
1. we get SIGKILLs from jobs that shouldn't use more than 2GB (this might arguably be a ghc issue).

The last run was successful and took a bit more than 47 minutes (the limit is 50).   I rebased to get rid of some of the more noisy attempts at getting travis to do its job and removed some stale package dependencies.  If the next run completes successfully, I will merge.

PR #302, opened 2015-09-17
Make agentRoles transaction return a list instead of a set

fhartwig:


fisx:
looks good.  if travis works from re-running it or from rebasing it to #301, you can merge this.

fhartwig:
Looks like the error was transient. Merging.

Issue #234, opened 2015-07-20
Prune stale dependencies from package files.


fisx:
The `Makefile` (on top level, not inside the packages) has a rule `%.packunused` that lists all unused dependencies in a cabal file.  However, it only looks at the modules listed in the cabal file, and some modules are only implicitly part of the package and not listed there (like everything collected by hspec-discover).

I can think of two ways to solve this:

1. hack something together in perl in the top-level Makefile that adds missing modules into the cabal files.  pseudo-code: ``perl -ne 'if (/-- @COLLECT_SPECS_START@/) { $bracket_open = 1; } else if (/-- @COLLECT_SPECS_END@/) { print (`find thentos-tests/src -name '*.hs'`) $bracket_open = 0; } else if ($bracket_open) { ... }' thentos-tests/thentos-tests.cabal``
2. consider https://github.com/sol/hpack.  If I got this right in the 4s I looked at it, it lets you write yaml files that compile to cabal files, but also insepct the package and add information like implicit modules.

I think 2. looks more promising.

fisx:
It would also be nice if an extra effect of resolving this issue would be that we have version bounds in all cabal files.  The rule should be: if `a.b.c.d` is installed, the bounds should be `>= a.b.c.d` and < `a.(b+1)`.

jkarni:
I worry a little about packunused. In particular, if we want to constraint the version of a transitive (indirect) dependency, that's not going to show up in our imports, so it'll also get flagged by packunused. But removing it would be wrong.

I guess we should commit to commenting on such dependencies, and then running running packunused locally rather than as part of the tests?

fisx:
> I worry a little about packunused. In particular, if we want to constraint the version of a transitive (indirect) 

This is a valid concern.  Comments would be an option, but it would be even nicer to have a machine-readable way of tagging these packages as  not-unused.

jkarni:
Assigning to myself since it's now part of #178.

jkarni:
We'd need https://github.com/sol/hpack/pull/36. Happy to contribute upstream for that, but this makes me think that for #178 we're better off just using `force-reinstall` for now.

fisx:
what's the status of this?

options:

- take it out of milestone 2 and wait for hspec to mature, or for somebody to be bored and get to work on this.
- close it as irrelevant

i would prefer to get done with this soon, as we have other fun stuff waiting for us.  @jkarni?

jkarni:
[...]

PR #308, opened 2015-09-22
reverting the extra source files in cabal attempt to fix test dependency tracking

robx:
It doesn't work (anymore? seems to me like since I upgraded stack).


fisx:
Possibly relevant: #234.

PR #316, opened 2015-09-28
Email change

### LINKED ISSUE #311: security of email and username change in thentos-core
If a logged-in user changes her email address, the new email address must not be confirmed, but it is.

Possible solutions:

1. set unconfirmed-bit in user and invalidate the session token.  this would kick the user out of the system and force her to activate the user immediately.  if something goes wrong, the user will stay invalid.  not ideal from a UX perspective.
2. keep the old email address until the new one is validated, and leave the user logged-in.  not ideal from a security perspective: if the user has lost control over an email address to an identity thief, this makes it hard to switch it off.

I still like 2 best from all the options I can think of.

jkarni:
Fixed #311

### DIFF
--- a/thentos-core/src/Thentos/Action.hs
+++ b/thentos-core/src/Thentos/Action.hs
@@ -30,7 +30,7 @@ module Thentos.Action
     , addPasswordResetToken
     , resetPassword
     , changePassword
-    , changePasswordUnconditionally
+    , _changePasswordUnconditionally
     , requestUserEmailChange
     , confirmUserEmailChange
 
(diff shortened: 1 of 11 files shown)


jkarni:
I *think* this is what we're supposed to do, but would love a nitpicking review :)

thentos-core/src/Thentos/Action.hs
```diff
@@ -278,18 +262,17 @@ changePassword uid old new = do
...
 
--- | Unconditionally change the password of a user.
---
 -- SECURITY: As a caller, you have to make sure that the user is properly authenticated and
--- authorized to change the password before calling this function!
+-- authorized to change the password before calling this function
 -- FIXME Reconsider/fix security model once our usage of LIO or a suitable replacement has
--- been clarified.
+-- been clarified
 changePasswordUnconditionally :: UserId -> UserPass -> Action e ()
```
fisx: At some point in the distant past I tried to introduce a naming convention for actions that should remain local to a module because they do not enforce any satisfactory lio policy.  I'll try that again now.

thentos-core/src/Thentos/Transaction.hs
```diff
@@ -192,48 +190,19 @@ confirmUserEmailChange timeout token = do
...
+-- | Change password. Should only be called once the old password has been
+-- verified.
+changePassword :: UserId -> HashedSecret UserPass -> ThentosQuery e ()
+changePassword uid newpass = do
     modified <- q
     case modified of
         1 -> return ()
         0 -> throwError NoSuchUser
-        _ -> impossible "updateUserField: unique constraint on id violated"
+        _ -> impossible "changePassword: unique constraint on id violated"
```
fisx: how do you know it's a unique constraint, and not something else?

jkarni: What else would it be? In this branch, more than 1 records were updated, which means the uniqueness constraint on `id` was violated.

fisx: ah.  right.  oh, and now i can also see the "impossible".  sorry.

fisx:
LGTM.  are my changes ok?

fisx:
seems to work.  @jkarni?

Issue #318, opened 2015-09-28
Add personas to DB


ChristianSi:
A persona belongs to a user and one or multiple services (= processes). A user may have at most one persona per process. (Note: the latter requirement may be relaxed in the future).

fisx:
[just adding my own, disorganized thoughts]

- I think this means service needs a field `uri_prefix` (or `uri`, but implicitly still meaning "prefix).  If a new process in created on a3, the corresponding new service is created.
- We need to communicate to a3 that services are identified by URI prefices, which means that all resources semantically living in a process need to have paths with its prefix.
- What would it mean for one service to have several prefices?  What if some prefices are shared by several services?  It may both be a bad idea, but perhaps it would be worth writing down a compelling reason why somewhere in the code.  (I suggesting to clutter the documentation, again.  Ignore me.  :-)

fisx:
A few more points [@ChristianSi, is this the right place to dump them?  If not, any suggestions?]

- Thentos needs to maintain individual adhocracy session tokens per persona, otherwise two personas can be linked when using the same session token.  This could be done by applying a suitable one-way function to the thentos session token (whould "hash thentos session token and persona" work?), or we could maintain yet another table.
- What do batch requests do to the idea of chosing a persona based on the request url?  Batch requests have many URLs, which may belong to different processes (at least in theory).  One solution would be to take batch requests apart in the proxy, and re-assemble the responses on the way back.  Not pretty, though.
- A3 has an admin role and a rest end-point where we can just put new / changed users and do not have to worry about the qualification process at all (even less than now?).  Not sure if we need it, but it may make things simpler.

ChristianSi:
@fisx I'd say it's not the right place, functional requirements must be specified as part of the user story, while tickets are about how to implement them. Tickets may request technical requirements, the the functional requirements must come from the story. Also, functional requirements shouldn't be added after a story is in process -- though in this case it's a bit of a problem that the story doesn't contain any explicit requirements ;)

Re 1: that's not a problem, since A3 never sees any session tokens, all the get from us as identifying information is the persona name (user name from A3's viewpoint).

Re the rest: let's talk that through...

ChristianSi:
Re 2 (after talking): It's unclear whether we need that at all (I'm very skeptical) and it's definitively not part of this story.

Re 3: that's not a requirement, but good to know -- might come handy.

fisx:
[...]

PR #331, opened 2015-10-07
Add personas and contexts to DB

ChristianSi:
Implements #318. Backend-code only, work in progress.


thentos-core/src/Thentos/Backend/Core.hs
```diff
@@ -133,6 +133,12 @@ thentosErrorInfo other e = f e
         (Nothing, err400, "thentos session not found")
     f NoSuchServiceSession =
         (Nothing, err400, "service session not found")
+    f NoSuchPersona =
+        (Nothing, err400, "persona not found")
+    f NoSuchContext =
+        (Nothing, err400, "context not found")
+    f MultiplePersonasPerContext =
+        (Nothing, err400, "Cannot register multiple personas by a user for the same context")
```
fisx: All other error messages start with a small letter.  And I would mention sybil attacks here:

"sybil attack prevention: cannot register multiple personas of the same user for the same context"

ChristianSi: Will be changed in the next commit. Nevertheless, "Sybil" is a proper name, therefore the error message will still start with an uppercase letter ;)

ChristianSi:
Should be OK for reviewing + merging, if somebody wants to do that. Tests are still incomplete but may be added in another PR.

thentos-core/src/Thentos/Types.hs
```diff
@@ -211,6 +215,38 @@ instance Aeson.FromJSON GroupNode where parseJSON = Aeson.gparseJson
...
+    deriving (Eq, Ord, Show, Read, FromJSON, ToJSON, Typeable, Generic, IsString, FromField,
+              ToField)
+
+newtype ContextDescription = ContextDescription { fromContextDescription :: ST }
+    deriving (Eq, Ord, Show, Read, FromJSON, ToJSON, Typeable, Generic, IsString, FromField,
+              ToField)
+
+data Context = Context
+  { _contextId           :: ContextId
+  , _contextOwnerService :: ServiceId
```
fisx: Rename to `_contextService`?

thentos-core/src/Thentos/Util.hs
```diff
@@ -86,16 +83,6 @@ readsPrecEnumBoundedShow _ s = f [minBound..]
         s' = show x
 
 
--- | Path concatenation for avoiding double slashes in paths.  One
```
fisx: Why was this moved?  And why the trivial extra sub-functions?  I would prefer to undo this; if you disagree, the sub-functions should at least be marked with the INLINE pragma (even though that may be happening automatically).

thentos-core/src/Thentos/Transaction.hs
```diff
@@ -268,6 +269,93 @@ unregisterUserFromService uid sid = void $
...
+-- Connect a persona with a context. Throws an error if the persona is already registered for the
+-- context or if the user has any *other* persona registered for the context. (As we currently
+-- allow only one persona per user and context.)
+registerPersonaWithContext :: Persona -> ContextId -> ThentosQuery e ()
+registerPersonaWithContext persona cxtId = do
+    -- Check that user has no registered personas yet
+    res <- queryT [sql| SELECT count(*)
+                        FROM personas pers, personas_per_context pc
+                        WHERE pers.id = pc.persona_id AND pers.uid = ? AND pc.context_id = ? |]
+                  (Only $ persona ^. personaUid)
```
fisx: I think the `context_id = ?` is too much?  Or how are there two `?` and only one argument?

fisx:
Is `Persona`:`process` 1:1, i.e. every context has *exactly* one persona?  Is the global context (empty path in `ProxyUri`) implicitly always there?  Or explicitly?

fisx:
What about `MultipleContextsPerPersona` in `registerPersonaWithContext`?

fisx:
I think I found a runtime error, which convinced me that you should probably complete the tests. (-:

thentos-core/src/Thentos/Util.hs
ChristianSi: I created the `stripLeadingSlash` and `stripTrailingSlash` helper functions because I need one of them in `renderProxyUri`. I cannot import Util into Types without an import loop and it felt strange to have the helpers in one module and the operator using them in another, hence I moved everything into Types.

As for the INLINE pragma, I would tend to let the compiler worry about that, except in cases where we found a performance bottleneck via benchmarking. Also, is there even a performance difference between a top-level definition and one defined in a where clause?

ChristianSi:
@fisx:

> Is `Persona`:`process` 1:1, i.e. every context has *exactly* one persona? 

It's 1:n, a user may use the same persona for multiple contexts. But...

> What about `MultipleContextsPerPersona` in `registerPersonaWithContext`?

... currently I don't allow a user to register multiple persons for the same context (Sybil attack), hence that error if they try.

> Is the global context (empty path in `ProxyUri`) implicitly always there?  Or explicitly?

That's up to the application to decide.

> I think I found a runtime error, which convinced me that you should probably complete the tests. (-:

Hey, what did you expect? SQL is dynamically typed, of course it doesn't work unless tested ;) But that's fine, I hadn't "officially" asked for a review and planned to complete the tests anyway. Will do that now.

thentos-core/src/Thentos/Transaction.hs
ChristianSi: Indeed, the context_id seems spurious.

thentos-core/src/Thentos/Types.hs
ChristianSi: I think `_contextOwnerService` is clearer but admittedly it's also a big longish. OK, I'll change it.

ChristianSi:
I completed the transaction tests now; should be good for a final review and possibly merge.

fisx:
looks good.

@ChristianSi could you check if you want to act on any of my last comments?  When done, you can merge.

ChristianSi:
@fisx:

> perhaps it would even be useful to have a function that takes a table name and a number of expected rows in that table and returns a `Spec`?

Good idea, I'll add that to the test utilities.

ChristianSi:
There now is a `rowCountShouldBe` test function and I also refactoring the existing test code to use it.

ChristianSi:
Travis failed with one of its random Selenium timeouts, merging anyway.

Issue #228, opened 2015-07-17
Port Snap frontend to Servant?


jkarni:

fisx:
I am very much in favour of this.  I think the way to do this is to look at the hello-world demo and replace our home-made protocol with oauth2-server (thentos-core) and oauth2-client (hello-world service).  It should be clear what needs to be done if you

1. run both helloworld and thentos-core
1. login on thentos as god/god and create a service
1. write the credentials to the helloworld-config and restart that
1. navigate to localhost:8000/ and take it from there

(currently buggy.)

fisx:
solved in #386

PR #335, opened 2015-10-11
DO NOT MERGE (removing snap)

fisx:
work on #228


fisx:
@jkarni: i think this can also be closed, right?

PR #337, opened 2015-10-14
Two refactorings to make transaction testing more convenient

ChristianSi:
* The helper functions for running ThentosQuery's from a pool have been
  renamed for clarity
* rowCountShouldBe (introduced yesterday) now reports as error location the
  place where it was called rather than its own implementation


jkarni:
LGTM (waiting for travis).

PR #338, opened 2015-10-14
DO NOT MERGE (lio)

fisx:
Here are a few ideas on how `lio` can be used more conveniently.

I will keep track of our off-band discussion here until we reach a conclusion.

### DIFF
--- a/thentos-core/src/LIO/Missing.hs
+++ b/thentos-core/src/LIO/Missing.hs
@@ -11,27 +11,30 @@
 module LIO.Missing
 where
 
-import Control.Exception (Exception)
 import LIO.Core (MonadLIO, liftLIO, taint, guardWrite)
+import LIO.Error (AnyLabelError)
 import LIO.Label (Label)
 import LIO.DCLabel (DCLabel, (%%))
 
 import qualified LIO.Exception as LE
 
 
-tryTaint :: (MonadLIO l m, Label l, Exception e) => l -> m r -> (e -> m r) -> m r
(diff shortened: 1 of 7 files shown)


ChristianSi:
I'm don't yet see how we're supposed to define safe actions, i.e. how to actually use LIO for authorization control.

fisx:
```haskell
safeFreshServiceId = makeActionSafe unsafeFreshServiceId
safeLookupConfirmedUser uid = withUserHasId uid $ unsafeLookupConfirmedUser uid
```

In practice, you would probably inline `unsafeLookupConfirmedUser` so you have one function for this action as before:

```haskell
lookupConfirmedUser :: UserId -> Action e (UserId, User)
lookupConfirmedUser uid = withUserHasId uid $
  _lookupUser $ T.lookupConfirmedUser uid
```

Does that help?

ChristianSi:
@fisx:

> Does that help?

Yes, looks nice enough.

How to combine multiple conditions (OR or AND)? And what happens if the condition is not fulfilled, does that depend on the `with...` function or can it be controlled by the action?

fisx:
I answered your two questions with two commits, in reverse order.

Re OR, AND: call `makeActionSafe'` with a predicate provided at the end of the module.

(sorry, i just push-forced.)

ChristianSi:
That's means for each predicate (e.g. `withUserHasRole`) we would need a `...Catch` alternative? That sounds a bit inflexible to me.

fisx:
of course you can just use `Action` as a `MonadError` and do things as originally intended...

fisx:
closing in favour of #341

PR #339, opened 2015-10-14
Define actions wrapping transactions for persons and contexts

### LINKED ISSUE #318: Add personas to DB

ChristianSi:
Authorization control is still work-in-progress and only noted in FIXMEs.

Fixes #318 (the rest of it, transactions were added in #331 already).


fisx:
@ChristianSi: If I don't hear from you first, I'm going to merge this later today unchanged, and you can process the comments next week elsewhere.

PR #341, opened 2015-10-16
Simplified Authorization API

fisx:
Based on a proposal by @ChristianSi and the earlier #338, this provides simplified access to lio-based authorization checks.

I demonstrate viability in (1) unit tests and (2) the actions for personas and contexts.

There is also independent cleanup code that I am happy to re-submit elsewhere should anything here are judged unfit for merging.

This pr puts us in a position to chose between two options:

1. drop dependency on lio and continue with the simplified api alone.
2. find use cases that make depending on lio the better choice.

This should happen elsewhere, too.

### DIFF
--- a/thentos-core/src/LIO/Missing.hs
+++ b/thentos-core/src/LIO/Missing.hs
@@ -33,6 +33,8 @@ tryGuardWrite label onSuccess onFailure = do
       Left e -> onFailure e
       Right () -> onSuccess
 
+-- | Test whether guard-write against a given label violates current clearance.  In other words:
+-- whether given label can flow to clearance.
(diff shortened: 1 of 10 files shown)


ChristianSi:
As long as we have no use cases where LIO is advantageous, I'm in favor of dropping it. We can still easily re-add it later should we run into such use cases.

ChristianSi:
I made some comments, otherwise this looks good! @fisx Feel free address them in whatever way you feel appropriate and then merge.

ChristianSi:
The renamed module still shows up somewhere and makes Travis unhappy.

PR #342, opened 2015-10-17
Makefile rules for sensei

fisx:


thentos-core/src-devel/Paths_thentos_core.hs
```diff
@@ -0,0 +1,6 @@
+-- | data files and development mode.  for details, see:
+-- http://neilmitchell.blogspot.de/2008/02/adding-data-files-using-cabal.html
+module Paths_thentos_core where
+
+getDataFileName :: FilePath -> IO FilePath
+getDataFileName = return
```
jkarni: What's the connection to the rest of the PR? And are we sure this is doing the right thing in non-development mode? I don't see how it'd be able to behave differently.

fisx: It is, because nobody knows about this directory except the sensei make rules.

The reason why it is needed is that I create one giant fake package out of all packages that are watched.  The fake package does not have this module, but it is imported somewhere.

Makefile
```diff
@@ -38,11 +41,58 @@ build-packunused:
...
+# sensei' via the SENSEI_ARGS shell variable.  see sensei and hspec
+# docs for details.
+
+sensei:
+	cabal sandbox hc-pkg -- unregister --force thentos-tests || true
+	cabal sandbox hc-pkg -- unregister --force thentos-core || true
+	cd thentos-tests && cabal clean
+	cd thentos-core && cabal clean
+	cabal exec -- sensei \
+	  -i./thentos-core/src/ \
```
jkarni: This is quite different than my `sensei` setup (when I used it, and if I do now, since I'd use `stack`). Is it best as part of the repo, rather than everyone having their own setup?

thentos-core/src/Thentos.hs
```diff
@@ -27,7 +28,7 @@ import Control.Concurrent.MVar (MVar, newMVar)
 import Control.Concurrent (ThreadId, threadDelay, forkIO)
 import Control.Exception (finally)
 import Control.Monad (void, when, forever)
-import Crypto.Random (ChaChaDRG, drgNew)
+import "cryptonite" Crypto.Random (ChaChaDRG, drgNew)
```
jkarni: Why is this needed now?

thentos-core/src-devel/Paths_thentos_core.hs
jkarni: Why not use the cabal macros file?

Makefile
```diff
@@ -38,11 +41,58 @@ build-packunused:
...
+	  -i./thentos-core/src-devel/ \
+	  -i./thentos-tests/src/ \
+	  -i./thentos-tests/tests/ ./thentos-tests/tests/Spec.hs $(SENSEI_ARGS)
```
jkarni: Without including `autogen/cabal_macros.h`, I don't think CPP will work (though we don't yet use it).

thentos-core/src-devel/Paths_thentos_core.hs
jkarni: ok

thentos-core/src/Thentos.hs
fisx: very good question.  my best guess is that this broke during a recent merge?

Makefile
fisx: I guess this question applies to the entire `Makefile`.  My argument for keeping it is that it easy to ignore entirely, and it contains a number of useful hacks that people can either use or mimic with their own tools.

Should we take a vote?

1. remove Makefile
2. keep Makefile, but do not add sensei rules
3. keep Makefile and add sensei rules.

I'm 3, but can live with all three.

fisx: @jkarni, you suggested to use .ghci instead of the `-i` cli arguments here. I think this is a bad idea, because it makes ghci crash with complaints about modules being present both in thentos-core and thentos-tests (this is what the unregister-commands are for).  In this make rule, the problem is explicit, well documented, and contained.  In .ghci, it will be harder to spot, and harder to work around.

ChristianSi:
@fisx:

> Should we take a vote?

I use "make hlint" regularly and I don't know a replacement for that, so I'm very much against (1).

(3) is best, I'd think.

Makefile
jkarni: OK

fisx:
Erm.  It doesn't get any prettier it seems, but CPP is working for me now.  Any suggestions?  I guess if we could generate just the macros header on the fly that would be nicer.

@fhartwig, did I miss your vote (no Makefile, old Makefile, Makefile plus sensei rules),  or do you not have an opinion?

fhartwig:
@fisx I don't have a strong opinion either way. I don't think having the Makefile is a problem, and we might as well add the sensei stuff.

PR #345, opened 2015-10-19
Services are now ownable only by real users, not other services.

### LINKED ISSUE #332: Make Service non-hierarchical.
About our earlier off-line discussion on ServiceId, hierarchical services, and agents:

1. I think giving serviceOwner type `UserId` instead of `Agent` makes sense.

1. Agent is also used as Session owner, which is a concept that is more extensively used in the thentos-as-a-server code.  We want to be able to, but not forced to, distinguish between services and users that connect to a rest api.  I don't think we should reomve the `Agent` type.

Does this make sense?  Any objections?

fisx:
Fixes #332

### DIFF
--- a/thentos-core/schema/schema.sql
+++ b/thentos-core/schema/schema.sql
@@ -34,12 +34,10 @@ CREATE TABLE IF NOT EXISTS user_roles (
 
 CREATE TABLE IF NOT EXISTS services (
     id            text       PRIMARY KEY,
-    owner_user    bigint     REFERENCES users (id),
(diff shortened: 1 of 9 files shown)


ChristianSi:
You forgot to adapt `schema.sql`: the `owner_service` column should be deleted and

    CHECK ((owner_user IS NULL) <> (owner_service IS NULL))

should be replaced by a simple NOT NULL constraint on the `owner_user` column.

Looks good otherwise.

PR #349, opened 2015-10-21
Group management

### LINKED ISSUE #319: Group management
Groups can be assigned to personas. Groups are (probably) just strings. There is a Role that is required for assigning and un-assigning groups. Personas and groups can be assigned to groups. Re-activate the function that flattens all groups to which a persona belongs (either directly or indirectly through another group). Cf. existing GroupNode data structure which might need a thorough rewrite.

ChristianSi:
Fixes #319.


fisx:
Looks good!

I removed the indicated dead types and found two small things.  Back to you, @ChristianSi.

PR #351, opened 2015-10-21
Servant combinators for digestive-functors

fisx:
This is a self-contained step on the way to decommissioning snap.

Missing (can be done elsewhere):
 - file upload
 - csrf protection


fisx:
(review should probably happen on the overall diff, not on the individual commits.)

PR #353, opened 2015-10-22
Remove snap

fisx:


fisx:
(will be back with more soon...)

PR #357, opened 2015-10-26
Create user in Thentos without relying on A3

### LINKED ISSUE #320: Change user activation process
Users are activated by Thentos, independently from A3. A confirmed user can only log into A3 using a persona. If the persona is used for the first time, a new A3 user is created that has the persona name as name. The user's email address is either empty (requires changes in A3) or set to some unique random values. Neither the actual name nor email of the user are ever exposed to A3.

Handle and store the mapping from A3 user names/paths to persona names. Note: For the beginning we'll probably work with a global A3 user management rather than process-specific users.

ChristianSi:
Now ready for review. Fixes #320.

Note that after this has been merged, the A3 config in the staging server should be changed as well by setting

        adhocracy.skip_registration_mail = true

in `etc/development.ini.in` (and running buildout).


fisx:
@ChristianSi I think that was the problem.  If there is another one please assign back to me.

fisx:
Re. configifier problem: Not sure if that helps, but the `Tagged` is only required around config trees for using, say, `>>.`.  In order to compose trees all sub-trees must be un-tagged.

fisx:
@ChristianS: I merged master with https://github.com/liqd/thentos/pull/376 in it.  Hope that's ok!

ChristianSi:
@fisx Thanks for your changes, they're all right with me.

fisx:
I made very few more changes (mostly comments).  @ChristianSi, please merge if you are ok with everything.  (has time until monday.)

PR #315, opened 2015-09-28
Cleanup: remove github sso fragments; some random details.

fisx:

Issue #328, opened 2015-10-05
testing purescript code


fisx:
Write one or two tests for the existing purescript functions.  Define a simple way to call the test suite (doesn't have to be integrated with the backend tests via selenium yet for now).

fhartwig:
# Current status

I initially wanted to use the test infrastructure that comes with the purescript tools (so we could just `pulp test` to run the tests). The problem with that is that this uses node.js to execute the tests, which doesn't work because we're using `xmlhttprequest` in our tests, which apparently doesn't exist in node. So we need to run the tests in the browser.

I've started using [purescript-spec](https://github.com/owickstrom/purescript-spec) which supports running tests either in the browser or with node. To use it in the browser, I wanted to use [https://github.com/owickstrom/purescript-spec-reporter-mocha], which seems to work well enough, except that it hangs indefinitely when the test does anything asynchronously. I'm not sure whether this is a bug or just me being incompetent with js stuff, but I've filed an issue on the project.

fisx:
Potentially relevant: https://github.com/slamdata/purescript-webdriver

fisx:
Ah, my mistake: purescript-webdriver only helps if one wants to run selenium from node.  Doesn't help with getting test suites in the browser to work.

PR #358, opened 2015-10-26
Introducing purescript

### LINKED ISSUE #325: Deliver login callback to browser
This ticket is mostly to get things going on the frontend, but it should also not be a detour on the way to completing the US.  Sub-tasks (in no particular order).

- Write a purescript callback that takes a user name and password, calls the login endpoint on one of our rest apis (adhocracy/simple would probably be most useful), and returns the generated session token or throws a js exception.
- Setup a preliminary build environment.
- Provide a servant endpoint to download the js code (it's ok if the function is no wrapped in a module yet, only touch the subject of amd or commonjs if it makes things easier).
- Provide an html page in which we can go into the debug mode and call the purescript callback by hand.
- Write tests and provide a way to run them.

fisx:
Contains a trivial (and not very well-written) purescript function that, when you run `thentos` and direct your browser to `/js` (in the backend) and open the console, shows you that it can talk to the backend.

Tests establish that the js link exists.  Missing: frontend-test suite.  See #328.

If you review this, you may want to read the total changes and not the indiviual commits.  I have taken quite a detour before putting purescript code mostly into its own directory, outside of any hackage packages.

Fixes #325


fisx:
(One issue is that travis now takes >30min again.  Can we tweak the caching?)

fisx:
Note to self: I think there is still something wrong with the way we call psc from cabal.  My best guess right now is that `--only-dep` changes something about order in which the hooks are called and the other events that occur in between.

It may be best to separate the two build-processes entirely.

thentos-purescript/src/Main.purs
```diff
@@ -0,0 +1,68 @@
...
+    then case eitherDecode res.response of
+      Right v -> Right v
+      Left e  -> Left $ DecodeError $ "error decoding response: " ++ show (Tuple res.response e)
+    else
+      Left $ ConnectionError $ "server responsed with error: " ++ show res.status
+
+main :: forall ajax err2 console. Eff (ajax :: AJAX, err :: EXCEPTION, console :: CONSOLE) Unit
+main = do
+  log "Hello sailor!"
+  runAff throwException print $ loginUser "god" "god"
```
ChristianSi: Hmm, in Haskell we use 4 chars for indentation, but large parts of this file use only 2, is that intentional? I'd prefer to use 4 chars in PureScript, too.

thentos-purescript/src/Main.purs
```diff
@@ -0,0 +1,68 @@
...
+
+
+defRq :: AffjaxRequest Unit
+defRq = defaultRequest { headers = [ContentType applicationJSON, Accept applicationJSON] }
+
+type Username = String
+type Password = String
+type ThentosSessionToken = String
+
+
```
ChristianSi: Spurious empty line (according to our hs coding guidelines).

thentos-tests/src/Thentos/Test/Core.hs
```diff
@@ -182,7 +182,7 @@ withBackend beConfig as = inTempDirectory . withBackend' beConfig as
 
 -- | Run a @hspec-wai@ @Session@ with the backend @Application@.
 withBackend' :: HttpConfig -> ActionState -> IO r -> IO r
-withBackend' beConfig as action =
+withBackend' beConfig as action = do
```
ChristianSi: That 'do' looks very spurious, but well...

thentos-tests/tests/Thentos/Backend/Api/PurescriptSpec.hs
```diff
@@ -0,0 +1,78 @@
...
+module Thentos.Backend.Api.PurescriptSpec (spec, tests)
+where
+
+import Control.Monad.State (liftIO)
+import Data.Configifier ((>>.))
+import Data.Proxy (Proxy(Proxy))
+import Data.String.Conversions (cs)
+import Network.Wai (Application)
+import Network.Wai.Test (simpleBody)
+import Servant.API ((:<|>)((:<|>)), (:>))
```
ChristianSi: `(:<|>)(..)` would be easier to read, but that's just a matter of taste.

thentos-tests/tests/Thentos/Backend/Api/PurescriptSpec.hs
```diff
@@ -0,0 +1,78 @@
...
+    -- the `purescript` field in config.
+    context "When path given in config" . with (defaultApp True) $ do
+        describe "/js/*.js" $ do
+            it "is available" $ do
+                request "GET" "/js/thentos.js" [] ""
+                    `shouldRespondWith` 200
+
+            it "has the right content type" $ do
+                resp <- request "GET" "/js/thentos.js" [] ""
+                liftIO $ cs (simpleBody resp) `shouldContain` ("PS[\"Main\"].main();" :: String)
```
ChristianSi: How does this test the content type?

ChristianSi:
I made some small notes, looks good otherwise!

ChristianSi:
The initial comment says "Fixes #315" but that's an already merged and pretty unrelated PR, so that sounds like a mistake...

thentos-purescript/src/Main.purs
fisx: no, but you are right, I should add more rules to the style guide.  (-:

thentos-tests/tests/Thentos/Backend/Api/PurescriptSpec.hs
fisx: the problem with import wildcards is that you can't search for type constructor imports any more.  it's not a big deal in this case because type name and constructor name are the same, but in general it may be annoying.

thentos-tests/tests/Thentos/Backend/Api/PurescriptSpec.hs
fisx: doesn't.  will fix.  thanks!

fisx:
(updated the fixed issue to #325.  thanks.)

ChristianSi:
Thanks, will merge if Travis is happy.

PR #375, opened 2015-10-29
Terminology doc

### LINKED ISSUE #366: Create document defining the user management terms
Such as role, group, permission, etc.

Should be reviewed by everybody interested in thentos / adhocracy teams

fisx:
Fixes #366.

### DIFF
--- a/docs/terminology.md
+++ b/docs/terminology.md
@@ -1,0 +1,42 @@
+# Thentos Terminology
+
+## Introduction
+
+Let's start with explaining the obvious.
+
+Users, names, roles, groups, persmissions, etc. are names for rather
+abstract concepts.  As it goes with abstraction, having a good
+intuition of their practical meaning can be hard.  Different software
+systems use different abstractions and thus use the same names with
+different meaning without always being aware of the differences.
+
+This document is attempt to establish common terminology and make it
+accessible for everybody who gets in contact with thentos.
+
+As https://github.com/liqd/adhocracy3.mercator is the first web app to
+use thentos, we are covering the terminology for that, too.  Other web
+apps will have different terms and concepts, and this document may be
+extended in the future to cover more.
+
+
+### User
+
+### Persona
+
+pseudonym?  nym?  nick?
+
+### Role
+
+### Permission
+
+### Group
+
+### Adhocracy3
+
+#### User
+
+#### Group
+
+#### Role
+
+#### Permission
--- a/docs/terminology.md
+++ b/docs/terminology.md
@@ -77,6 +77,9 @@ FIXME: not sure if nyms are the right tool to achieve this moderator /
 contributor double-role distinctions.  perhaps roles alone can do that
 when exposed to the UI?
 
+FIXME: see also:
+https://cms.liqd.net/intranet/copy_of_projekte/adhocracy-3/grundkonzept-und-terminologie#section-15
+
 
 ### Role
 
@@ -106,22 +109,22 @@ groups.)
 
 ### User (Adhocracy3)
 
-See https://cms.liqd.net/intranet/copy_of_projekte/adhocracy-3/grundkonzept-und-terminologie.
+See https://cms.liqd.net/intranet/copy_of_projekte/adhocracy-3/grundkonzept-und-terminologie#section-14
 
 
 ### Group (Adhocracy3)
 
-See https://cms.liqd.net/intranet/copy_of_projekte/adhocracy-3/grundkonzept-und-terminologie.
+See https://cms.liqd.net/intranet/copy_of_projekte/adhocracy-3/grundkonzept-und-terminologie#section-16
 
 
 ### Process Role (Adhocracy3)
 
-See https://cms.liqd.net/intranet/copy_of_projekte/adhocracy-3/grundkonzept-und-terminologie.
+See https://cms.liqd.net/intranet/copy_of_projekte/adhocracy-3/grundkonzept-und-terminologie#section-17
 
 
 ### Permission (Adhocracy3)
 
-See https://cms.liqd.net/intranet/copy_of_projekte/adhocracy-3/grundkonzept-und-terminologie.
+See https://cms.liqd.net/intranet/copy_of_projekte/adhocracy-3/grundkonzept-und-terminologie#section-19.
 
 
 ## FIXME:
--- a/docs/terminology.md
+++ b/docs/terminology.md
@@ -15,7 +15,7 @@ accessible for everybody who gets in contact with thentos.
 
 Unless stated otherwise, all terms relate to concepts as implemented
 in thentos.  Having that said, as
-[Adhocracy 3](https://github.com/liqd/adhocracy3) is the first web app to
+[Adhocracy3](https://github.com/liqd/adhocracy3) is the first web app to
 use thentos, we are covering the terminology for that, too.  Other web
 apps will have different terms and concepts, and this document may be
 extended in the future to cover more of them.
--- a/docs/terminology.md
+++ b/docs/terminology.md
@@ -13,30 +13,115 @@ different meaning without always being aware of the differences.
 This document is attempt to establish common terminology and make it
 accessible for everybody who gets in contact with thentos.
 
-As https://github.com/liqd/adhocracy3.mercator is the first web app to
+Unless stated otherwise, all terms relate to concepts as implemented
+in thentos.  Having that said, as
+https://github.com/liqd/adhocracy3.mercator is the first web app to
 use thentos, we are covering the terminology for that, too.  Other web
 apps will have different terms and concepts, and this document may be
-extended in the future to cover more.
+extended in the future to cover more of them.
 
 
 ### User
 
-### Persona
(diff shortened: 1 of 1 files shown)


docs/terminology.md
```diff
@@ -0,0 +1,143 @@
...
+intuition of their practical meaning can be hard.  Different software
+systems use different abstractions and thus use the same names with
+different meaning without always being aware of the differences.
+
+This document is attempt to establish common terminology and make it
+accessible for everybody who gets in contact with thentos.
+
+Unless stated otherwise, all terms relate to concepts as implemented
+in thentos.  Having that said, as
+https://github.com/liqd/adhocracy3.mercator is the first web app to
```
ChristianSi: I'd change that to: [Adhocracy 3](https://github.com/liqd/adhocracy3]

docs/terminology.md
```diff
@@ -0,0 +1,143 @@
...
+contributor double-role distinctions.  perhaps roles alone can do that
+when exposed to the UI?
+
+FIXME: see also:
+https://cms.liqd.net/intranet/copy_of_projekte/adhocracy-3/grundkonzept-und-terminologie#section-15
+
+
+### Role
+
+A role is a set that contains users and other roles.
```
ChristianSi: What's the purpose of a role and how it is used?

docs/terminology.md
```diff
@@ -0,0 +1,143 @@
...
+
+
+### Permission
+
+A permission is just a bit that tells you if an action is allowed or
+not.  A permission can be assigned to a user or a role.
+
+If permission X is assigned to role Q, then all users member in role Q
+are allowed to perform action X.
```
ChristianSi: Sounds like every permissions corresponds to *one* action, which is probably not intended.

docs/terminology.md
```diff
@@ -0,0 +1,143 @@
...
+
+
+### Group
+
+Groups are very similar to roles.  In fact while they still live
+inside thentos, they are the equivalent of roles in web apps.  Where
```
ChristianSi: To maybe rename role to "user role" and group to "persona role" for clarity?

docs/terminology.md
```diff
@@ -0,0 +1,143 @@
...
+app registered with thentos.  (If thentos only serves one web app,
+this is not much of a distinction.  But if there are many web apps
+sharing the same thentos server, each of them can manage their own
+groups.)
+
+
+### User (Adhocracy3)
+
+Like thentos user.  See
+https://cms.liqd.net/intranet/copy_of_projekte/adhocracy-3/grundkonzept-und-terminologie#section-14
```
ChristianSi: But corresponds to a persona from the viewpoint of Thentos!

docs/terminology.md
fisx: we could do that.  note that persona roles also differ from user roles by being local to services, but that's not a reason against the renaming.

docs/terminology.md
```diff
@@ -0,0 +1,143 @@
...
+There are two related use cases:
+
+*Pseudonymity:* Consider a discussion forum web app.  In some
+contributions, a user may want to use her real name to leverage or
+increase her reputation in the community.  In others, possibly even
+under the same discussion, she may want to remain anonymous.  This can
+be achieved by using two nyms: one identical to the user's real name,
+and one generated at random.
+
+*The Moderator who also wants to engage in a debate:* If the same user
```
ChristianSi: This should be generalized.

docs/terminology.md
```diff
@@ -0,0 +1,143 @@
...
+app, and having them assigned different roles (see below).
+
+FIXME: not sure if nyms are the right tool to achieve this moderator /
+contributor double-role distinctions.  perhaps roles alone can do that
+when exposed to the UI?
+
+FIXME: see also:
+https://cms.liqd.net/intranet/copy_of_projekte/adhocracy-3/grundkonzept-und-terminologie#section-15
+
+
```
ChristianSi: Services and contexts should be explained, too.

docs/terminology.md
```diff
@@ -0,0 +1,143 @@
...
+## FIXME:
+
+should we make this an alphabetical list?
+
+what about these terms?:
+
+- authorization error
+- authorization
+- identification
+- qualification (in the sense of berlin servicekonto)
```
fisx: also missing: "context".  a context is a path into a web app (e.g., a process or a proposal).  we may want to map personas and context onto each other, which blurs the line between app data and user data.  It is necessary to maintain the mapping inside thentos, because the mapping requires the UID, which to hide from the app is one of the core ideas of personas.

docs/terminology.md
```diff
@@ -0,0 +1,143 @@
...
+### Nym
+
+FIXME: used to be called persona in thentos, too.  fisx likes 'nym'
+better: less overloading, shorter, prettier.  should we have a vote on
+renaming it?  other options: pseudonym, nick, identity.
+
+A person may want to assume different identities at different times.
+A simple way to achieve that would be to create many users, but on the
+other hand some things like residential address or system
+notifications, should be shared between those identities.
```
fisx: add a sentence along the lines of: "this is a technical concept.  there are different UI alternatives, e.g., ..., of which the technical concept is independent."

docs/terminology.md
```diff
@@ -0,0 +1,143 @@
...
+Interactions between the real person and thentos or a web app are
+called actions.  An attempted action may trigger an authorization
+error (see below).
+
+
+### Nym
+
+FIXME: used to be called persona in thentos, too.  fisx likes 'nym'
+better: less overloading, shorter, prettier.  should we have a vote on
+renaming it?  other options: pseudonym, nick, identity.
```
fisx: change 'Nym' back to 'Persona'.

docs/terminology.md
fisx: motivation: roles are used for giving permissions to many users at once.  roles of roles are useful for example: user, confirmed user, admin.  an admin is a confirmed user with extra stuff, and we only need to talk about the extra stuff.

docs/terminology.md
fisx: final decision for now: change 'role' to 'user group' and 'group' to 'persona group'.

docs/terminology.md
ChristianSi: Generally, a permission allows a set of actions, not just one.

fisx: we don't have permission as a concept now, but authorization check and (perhaps) information flow control.

fisx: also needed as term: action.

docs/terminology.md
ChristianSi: We should also assign that persona groups are supposed to be sent to the service. So the service knows to which groups a persona allows and will accordingly allow or deny them to do specific things.

docs/terminology.md
```diff
@@ -0,0 +1,143 @@
...
+valid in a thentos installation globally, a group is local to a web
+app registered with thentos.  (If thentos only serves one web app,
+this is not much of a distinction.  But if there are many web apps
+sharing the same thentos server, each of them can manage their own
+groups.)
+
+
+### User (Adhocracy3)
+
+Like thentos user.  See
```
fisx: thentos persona.

docs/terminology.md
```diff
@@ -0,0 +1,143 @@
...
+### Group (Adhocracy3)
+
+Like thentos role.  See
+https://cms.liqd.net/intranet/copy_of_projekte/adhocracy-3/grundkonzept-und-terminologie#section-16
+
+
+### Process Role (Adhocracy3)
+
+Time-aware set of adhocracy permissions.  See
+https://cms.liqd.net/intranet/copy_of_projekte/adhocracy-3/grundkonzept-und-terminologie#section-17
```
fisx: example: "this group has that permission in process phase somethingorother."  thentos doesn't need to worry about adhocracy roles.

docs/terminology.md
```diff
@@ -0,0 +1,143 @@
...
+and one generated at random.
+
+*The Moderator who also wants to engage in a debate:* If the same user
+enters a channel where she acts as moderator, she wants to have
+different controls in the UI from ordinary users (e.g., she doesn't
+want to be able to accidentally write a comment, but she needs to be
+able to delete comments of others).
+
+This can be achieved by registering several nyms with the same web
+app, and having them assigned different roles (see below).
```
ChristianSi: We should probably also mention that sockpuppetry is another obvious use case which we DON'T want to support (but rather to preserve).

docs/terminology.md
fisx: @ChristianSi: i am changing this to:

> *Doing different things in the same place:*

with my next commit (coming up), could you check if you are happy, and if not, make a counter-proposal?

fisx:
Just added a few more terms.  I hope it's not getting too long?

@ChristianSi, @fhartwig: FYI; @rittermo: assigning to you; let me know if need to to something else.

thanks-

ChristianSi:
Looks pretty good now, I made some more comments re the new commits.

fisx:
Thanks @ChristianSi for the thorough review.  I processed most of it, and commented on the rest.

PR #376, opened 2015-10-29
Introduce purescript-halogen

fisx:
Table of contents:

- introduce purescript-halogen dependency.
- simple login indicator widget (with mock app logic) to demonstrate embedding into adhocracy.
- simple counter widget to demonstrate interaction between widget and surrounding app.
- two demos: counter app; iframes.


fisx:
See also: https://github.com/liqd/adhocracy3.mercator/pull/1816

thentos-adhocracy/src/Thentos/Adhocracy3/Backend/Api/Simple.hs
```diff
@@ -375,6 +378,7 @@ thentosApi actionState = enter (enterAction actionState a3ActionErrorToServantEr
 api :: Client.Manager -> AC.ActionState -> Server Api
 api manager actionState =
        thentosApi actionState
+  :<|> Thentos.Backend.Api.Purescript.api (Just "/home/mf/thentos/thentos-purescript/static/")
```
fisx: Confusingly, this is fixed in #377.

I suggest we merge #377 into master, then master into #376, then fix this by feeding `api` the `ThentosConfig` from `actionState`?

fisx:
two open issues:

1. build error on travis (and locally)
2. pulp doesn't propagate build errors to calling script.

both should be addressed before this can be merged.

fisx:
1. is fixed.

fhartwig:
Looks good, and it works for me now. I'll make a PR for pulp to propagate errors from psc, but we should not block this PR on that.

PR #379, opened 2015-11-11
Tweak purescript build script

fisx:


fisx:
(oops, just realized i called the branch "something-something-*buildout*@.  never mind.  (-:

fisx:
23e9a75 crashes on travis as intended; b357c89 should go through.  if it does, this pr is ready for review.

fhartwig:
Looks good, I'm just waiting for travis.

fisx:
(it looked good, but it was broken.  this seems more fixed...  sorry!)

fisx:
The tests all failed because a brittle test case in the server code.  It is easy to see from the time at which travis fails in the resp. commits that it now correctly catches psc errors and compiles if there are none.

The next commit will fix the offending test case.

fhartwig:
(Travis still failing, will push a fix in a minute).

Issue #39, opened 2015-04-15
Use SafeHaskell


mf59816:
SafeHaskell tracks calls to not explicitly trusted, potentially harmful functions.  SafeHaskell can be used to identify all calls to potentially harmful functions so we can decide whether we want to mark them as intentionally risky or fix them.

It has its false positives and negatives, the former can be marked explicitly as "don't care", and the latter only constitute a limination of, not a reason against using SafeHaskell.

jkarni:
Can we discuss a little the threat model? We'd have to mark a lot of imports as trusted, so in that sense we wouldn't get much benefit from `-XSafe` for those packages. 
(We'd also have to trust any of our own modules that use TH, since that can't be `-XSafe`.)

PR #385, opened 2015-11-16
Make captcha state storable / gc-able

ChristianSi:
Fixes [Taiga task #39](https://tree.taiga.io/project/fisx-thentos/task/39) (can I autolink that somehow?).

Contributions:

* `Thentos.Sybil.Captcha` module: a placeholder for functionality to come
* `makeCaptcha` action generates and returns a captcha and stores the
  proper solution in the DB
* `solveCaptcha` action checks whether a proposed solution is correct and
  deletes the solution from the DB (there is always only one attempt to get
  it right)
* Captchas added to the DB and still not solved are deleted as per the
  `captcha_expiration` config param (by default, after 1 hour)


thentos-core/src/Thentos/Types/Opaque.hs
```diff
@@ -0,0 +1,24 @@
+-- | Types whose internal representation is not accessible.
```
fhartwig: Is this module just a workaround for the missing export list in `Types.hs`. If so, I'd prefer to have an export list in `Types.hs` and move these types there.

thentos-core/src/Thentos/Sybil/Captcha.hs
```diff
@@ -0,0 +1,15 @@
+{-# LANGUAGE OverloadedStrings    #-}
+
+module Thentos.Sybil.Captcha where
```
fhartwig: I think `Thentos.Sybil.Captcha` should just be `Thentos.Captcha`, unless there is a point to the extra indirection that I'm missing.

fhartwig:
This looks good, modulo my two comments.

thentos-core/src/Thentos/Types/Opaque.hs
fisx: :+1:

everybody should have one.  it's just nobody has had the patience to write them.  (and at some point in the past i argued that things were still moving too much for it to be practical.  not the case any more!)

thentos-core/src/Thentos/Sybil/Captcha.hs
fisx: `Thentos.Sybil.Blacklists`, `Thentos.Sybil.Analytics`.  I think it should stay here.

fisx:
looks good, modulo several more comments.

fisx:
travis sais:

```
[14 of 33] Compiling Thentos.Transaction ( src/Thentos/Transaction.hs, dist/dist-sandbox-d106cceb/build/Thentos/Transaction.o )
src/Thentos/Transaction.hs:250:46:
    Not in scope: type constructor or class ‘Uri’
cabal: Error: some packages failed to install:
thentos-adhocracy-0.0.1.1 depends on thentos-core-0.0.1.1 which failed to
install.
thentos-core-0.0.1.1 failed during the building phase. The exception was:
ExitFailure 1
thentos-tests-0.0.1.1 depends on thentos-core-0.0.1.1 which failed to install.
The command "./misc/thentos-install.sh -c "--force-reinstalls"" failed and exited with 1 during .
```

but that doesn't make any sense.  line 250 is this:

```
        _ -> impossible "deleteService: multiple results"
```

and the entire module sais nothing about `Uri`.  i"ll just try once more.

fisx:
Ok, I still don't know how that explains the error message, but I'm glad you think you do.  :-)

Btw, the new error may be due to me rebasing the branch the servant submodule is sitting on.  There are limits to when it is ok to rebase, I guess.  I think we should update the submodule to the upstream master again and pluck the servant-session package and keep it as a local subdirectory for now.  That would make it easier to pull in new commits as they are merged.

ChristianSi:
@fisx Travis merges the branch into master before running the tests, hence the error: the `Types` module now has an export list, but `Uri` and friends aren't in it.

As for the submodule stuff, I've no idea what to do, but, for the record: I'm generally against rebasing! ;)

fisx:
Yes, rebase is bad.  Also, submodules are.  I tried a number of things to update the submodule, and I'm not sure what will happen to you after you pull.  What worked for me:

```bash
git rm submodules/servant
git submodule add -b <branch> <repo>
```

what may work for you:

```bash
git submodule deinit
git submodule sync
git submodule update --init
```

the `deinit` may be unnecessary.

fisx:
(Oh, and the errors are not fixed yet.  But now the branch contains the code that has them.)

PR #382, opened 2015-11-14
No snap -- DO NOT MERGE

fisx:
This is a major refactoring to replace the snap dependency with servant in the frontend.  The commits are not self-contained, but are hoped to provide a little bit of structure for the review.

This is ready for review, but is based on the branch of #357 and should not be merged before that.

PR #386, opened 2015-11-16
Replace snap with servant for core frontend.

fisx:
This is a major refactoring to replace the snap dependency with servant in the frontend. The commits are not self-contained, but are hoped to provide a little bit of structure for the review.

This obsolete #382.  The reason for the new PR is that this branch is re-based to a new branch (master).


ChristianSi:
Looks good, minus some comments I made. @fisx You don't have to address all of them (of course), but please check where it makes sense. OK to merge afterwards.

PR #387, opened 2015-11-19
Register widget; client code generation

fisx:
Commits should hopefully explain themselves.  Note that some of the committed purescript files are generated, and updates do not need to be reviewed.

The changes to `dev-howtos.md` are a bit noisy.  I am open to discuss how to improve that.

### DIFF
--- a/thentos-purescript/src/Register.purs
+++ b/thentos-purescript/src/Register.purs
@@ -118,11 +118,11 @@ body st = case Tuple st.stLoggedIn st.stRegSuccess of
         [ H.h2_ [translate "TR__REGISTER_SUCCESS"]
         , H.p_ [translate "TR__REGISTRATION_CALL_FOR_ACTIVATION"]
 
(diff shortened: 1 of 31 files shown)


fhartwig:
We should freeze purescript-halogen (and other dependencies?)

thentos-purescript/src/Register.purs
```diff
@@ -0,0 +1,207 @@
...
+eval (UpdateTermsAndConds newVal next) = do
+    modify (\st -> st { stTermsAndConds = newVal })
+    modify checkState
+    pure next
+
+checkState :: State -> State
+checkState st = st { stErrors = passwordMismatch st ++ invalidEmail st }
+  where
+    passwordMismatch :: State -> Array String
+    passwordMismatch st = if st.stPass1 == st.stPass2 then [] else ["passwords must match"]
```
fhartwig: (Nitpicking, feel free to ignore) Why not make `passwordMismatch` just an `Array String` and use the `st` from the outer scope?

PR #388, opened 2015-11-19
Captcha: construct challenge/solution pairs

fisx:
This is a port from hs-captcha.

It completes `Thentos.Sybil.Captcha` up to the point where an image is generated, but the distort function is still mostly missing.  https://hackage.haskell.org/package/friday could be a better fit for this task than https://hackage.haskell.org/package/diagrams (thanks @fhartwig).

I would like merge this as is and resume work later, because with these changes it will be possible to build and demonstrate a UI.


fisx:
(The problem that makes travis fail should be fixed in #387.)

fhartwig:
:+1: Looks good. Feel free to merge once travis passes.

PR #390, opened 2015-11-23
Remove workaround for pulp error reporting issues

fhartwig:


fisx:
LGTM.  I'll wait for travis and then merge.

fisx:
browserify is unhappy, but I don't understand why.

fisx:
I just checked locally both before and after a merge with master: thentos-purescript builds without problems on pulp-4.4.1.  On pulp-5.0.0, both with psc-0.7.5.3 and psc-0.7.6.1, there is a type error in LoginIndicator pointing to a change in the api of the purescript-generics package.

If travis is having the same error, then my guess would be that the previous error was due to a combination of version that none of us tried out locally?  And if it still has the browserify error, I still don't understand.

fisx:
Ah.  Yet another problem.  Ok, trying again...

fisx:
this is blocking me at #391, so i'll look into this now.

fisx:
what i've learned so far:

- `pulp browserify` crashes with a run-time js exception between writing *Browserifying...* and *Building...* to stdout.
- only on travis, not on my machine.
- versions of pulp, browserify, psc, are the same on travis and on my machine.
- `pulp browserify` and `pulp build` are both affected, so it is probably not a browserify issue.
- if i remove the `-O` flag for `browserify` again and downgrade pulp to 4.4.1, the problem goes away.

giving up for today...

fisx:
(the most recent travis error is about purescript-generics being too old.  upgrade helped on #391.)

ChristianSi:
Is this PR still relevant?

fisx:
I would like to leave it open because it is performance-relevant.  Labelled it low-prio.

fisx:
Word on the street is that gulp is a good replacement for pulp: http://www.arow.info/blog/posts/2015-12-17-purescript-intro.html#fnref1

This seems to be consistent with our experience so far.

fisx:
Closing in favour of #435.

PR #394, opened 2015-11-25
Integrate captcha into user registration workflow

ChristianSi:
This resolves Task 41: change registration form validation to check for captcha.

The core API now has "/user/captcha", "/user/register", and "/user/activate" endpoints for creating a user who must be able to solve the captcha correctly. The tests are the best documentation of the workflow.

Suitable frontend (widget) is still missing and handled in a different task. Two other things are missing and will be addressed in different cases:

* If the user name is not unique, the user currently has to solve a new captcha instead of just entering a new name. There is a FIXME in one of the tests for this -- I propose to fix it in Task 42: handle the case that user enters wrong data.
* The "register" endpoint doesn't send out any confirmation emails yet, so the complete workflow doesn't currently work. This will be addressed by Task 60: Modify addUnconfirmedUser action to send emails.

### DIFF
--- a/docs/dev-howtos.md
+++ b/docs/dev-howtos.md
@@ -3,6 +3,20 @@
 This document describes some of our development practices. Please help it
 grow!
 
+## Source comment keywords
+
+We use the following greppable keywords in the source code with the
+resp.  specific meaning:
+
+- *TODO*: note to self to developer currently working on this piece of
+   code.  to be fixed before merge.  should never appear in master.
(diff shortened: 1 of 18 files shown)


fisx:
Looks good.  Instead of commenting, I added everything I had to say in the form of commits.

I am going to merge this if travis has no further objections, but could you take a look at my changes and see if you can agree with everything?

Thanks!

PR #395, opened 2015-11-26
A3 Compat DO NOT MERGE

fhartwig:
My current idea how to solve https://tree.taiga.io/project/fisx-thentos/task/37 is that instead of putting the entire user resource URL in the `X-User-Path` header, we can just put the path component of the header. I've tried this on an older version of thentos and it seems to fix the problem. I haven't managed to test this with this version of thentos, though, since the email sending functionality seems to be broken, so I cannot create a user to log in with. (thentos crashes with `ActionErrorUnknown sendUserConfirmationMail: no such context`, where the argument to `context` seems to be `frontend_url`.

### DIFF
--- a/thentos-adhocracy/src/Thentos/Adhocracy3/Backend/Api/Simple.hs
+++ b/thentos-adhocracy/src/Thentos/Adhocracy3/Backend/Api/Simple.hs
@@ -99,6 +99,8 @@ import qualified Thentos.Backend.Api.Simple ()
 
 -- ** basics
 
+-- | This is a "Path" meaning an descriptor of an a3 Ressource, not a file
+-- system path or part of an URL
 newtype Path = Path { fromPath :: ST }
   deriving (Eq, Ord, Show, Read, Typeable, Generic, FromJSON, ToJSON)
 
@@ -442,7 +444,7 @@ activate ar@(ActivationRequest confToken) = AC.logIfError'P $ do
     pure $ RequestSuccess (Path . cs . renderUri $ externalUrl) stok
 
 -- | Make user path relative to our exposed URL instead of the proxied A3 backend URL.  Only works
--- for @/principlas/users/...@.  (Returns exposed url.)
+-- for @/principals/users/...@.  (Returns exposed url.)
 makeExternalUrl :: PersonaName -> A3Action Uri
 makeExternalUrl pn = createUserInA3'P pn >>= f
   where
@@ -453,7 +455,7 @@ makeExternalUrl pn = createUserInA3'P pn >>= f
         | otherwise = do
             config <- AC.getConfig'P
             let (Path fullPath) = a3backendPath config localPath
-            case parseUri $ cs fullPath  of
+            case parseUri $ cs fullPath of
                 Left err  -> throwError . OtherError $ A3UriParseError err
                 Right uri -> pure uri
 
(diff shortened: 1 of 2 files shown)


fisx:
I think this is your bug:

```sh
$ git grep frontend_url
[...]
thentos-core/src/Thentos/Config.hs:  :*> \
  Just "Please go to {{frontend_url}}{{activation_path}} to confirm your account."
```

Regarding changing the `X-User-Path` header: This is likely to have deep implications in adhocracy.  I think for now we need to deal with absolute URIs both towards frontend and towards backend.  If you want to do something else, we should discuss this before you spend too much time on it.

fhartwig:
Can you be more specific about the implications? As far as I can tell, the `X-User-Path` is only used for authentication in a3, and it seems like that handles both formats fine. I don't want to change anything that we return to the frontend. I've tested this now (the email issue was fixed in #396) and at least in my setup it seems to work. If you have any pointers where potential problems might be, that would be helpful.

thentos-adhocracy/src/Thentos/Adhocracy3/Backend/Api/Simple.hs
```diff
@@ -612,17 +614,18 @@ a3RenderUser uid _ = externalUrlOfDefaultPersona uid
...
+    -- FIXME: get rid of the fromJust, a3-Prefix should not be optional
+    a3Prefix = fromJust $ config >>. (Proxy :: Proxy '["a3-prefix"])
 
 userIdFromPath :: MonadError (ThentosError e) m => Path -> m UserId
 userIdFromPath (Path s) = do
     uri <- either (const . throwError . MalformedUserPath $ s) return $
         URI.parseURI URI.laxURIParserOptions $ cs s
     rawId <- maybe (throwError $ MalformedUserPath s) return $
+        -- FIXME: I don't think this will work, as the path may be something
+        --  like /foo/principals/users/ (see https://github.com/liqd/adhocracy3/pull/1849)
```
fisx: Then this should be fixed immediately, not left in as a FIXME, right?

fisx: (I mean: it is a BUG, not a FIXME.)

thentos-adhocracy/src/Thentos/Adhocracy3/Backend/Api/Simple.hs
```diff
@@ -612,17 +614,18 @@ a3RenderUser uid _ = externalUrlOfDefaultPersona uid
 -- | Convert a local file name into a absolute path relative to the A3 backend endpoint.  (Returns
 -- exposed url.)
 a3backendPath :: ThentosConfig -> ST -> Path
-a3backendPath config localPath = Path $ cs (exposeUrl beHttp) <//> localPath
+a3backendPath config localPath = Path $ a3Prefix <//> localPath
   where
-    beHttp     = case config >>. (Proxy :: Proxy '["backend"]) of
-                     Nothing -> error "a3backendPath: backend not configured!"
-                     Just v -> Tagged v
+    -- FIXME: get rid of the fromJust, a3-Prefix should not be optional
```
fisx: TODO, not FIXME.

thentos-adhocracy/src/Thentos/Adhocracy3/Backend/Api/Simple.hs
fhartwig: Yes, I think this should be fixed soon, but it's out of scope for this PR. I'll mark it as BUG and open a ticket.

fhartwig:
Closing in favour of #412

PR #396, opened 2015-11-30
Fix default verification mail template

fhartwig:
I think this slipped through in https://github.com/liqd/thentos/pull/357 . @ChristianSi , could you review this?


ChristianSi:
Looks good (still waiting for Travis).

PR #398, opened 2015-12-01
Modify addUnconfirmedUser action to send emails

ChristianSi:
Registration emails are now sent by the addUnconfirmedUser action instead of by its callers. Hence all interfaces (core/REST, core/HTML, a3) will behave in the same way and there is no duplication of functionality.

Resolves Task 60: Modify addUnconfirmedUser action to send emails.

### DIFF
--- a/devel.config
+++ b/devel.config
@@ -33,3 +33,29 @@ database:
 
 purescript:
     "./thentos-purescript/static/"
+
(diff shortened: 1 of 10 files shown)


fhartwig:
Looks good, just waiting for travis. :+1:

PR #412, opened 2015-12-03
Make thentos stick to User-Path format expected by A3

fhartwig:
Cleaned-up version of #395 .
Notes/questions for the reviewer:

* a3-prefix should clearly not be in thentos-core, since it is specific to `thentos-adhocracy`. I think thentos-adhocracy should work with a `ThentosA3Config` or something, that extends `ThentosConfig` instead of using `ThentosConfig` directly. However, I think making ThentosConfig extensible is probably not trivial enough to put it in this pr (since all the functions dealing with Config are monomorphic).
* In a discussion with @fisx we talked about only storing the path component of a user-path (i.e. not the entire URL) and construct the user resource path dynamically. I don't think this is a good idea anymore, since a3 also stores the user path on user activation, and we probably don't want Thentos' and a3's view of the user path to diverge.

### DIFF
--- a/thentos-adhocracy/src/Thentos/Adhocracy3/Backend/Api/Simple.hs
+++ b/thentos-adhocracy/src/Thentos/Adhocracy3/Backend/Api/Simple.hs
@@ -622,8 +622,9 @@ userIdFromPath (Path s) = do
     uri <- either (const . throwError . MalformedUserPath $ s) return $
         URI.parseURI URI.laxURIParserOptions $ cs s
     rawId <- maybe (throwError $ MalformedUserPath s) return $
-        -- FIXME: I don't think this will work, as the path may be something
+        -- BUG (I think) : I don't think this will work, as the path may be something
         --  like /foo/principals/users/ (see https://github.com/liqd/adhocracy3/pull/1849)
+        -- github issue at https://github.com/liqd/thentos/issues/411
         stripPrefix "/principals/users/" $ dropWhileEnd (== '/') (cs $ URI.uriPath uri)
     maybe (throwError NoSuchUser) (return . UserId) $ readMay rawId
 
--- a/thentos-adhocracy/src/Thentos/Adhocracy3/Backend/Api/Simple.hs
+++ b/thentos-adhocracy/src/Thentos/Adhocracy3/Backend/Api/Simple.hs
@@ -39,7 +39,6 @@ module Thentos.Adhocracy3.Backend.Api.Simple
     , thentosApi
     ) where
 
-import Data.Maybe (fromJust)
 import Control.Lens ((^.), (&), (<>~))
 import Control.Monad.Except (MonadError, throwError)
 import Control.Monad (when, mzero)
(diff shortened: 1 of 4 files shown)


fisx:
Re. polymorphic configifier types: yes, please leave that for a separate PR.

Re. path vs. URL: What does the code currently do?

fisx:
**Note to self:** the failing test case requires the test config in order to work.  I suggest to

1. Change the type of `Thentos.Test.Core.withBackend` to `HttpConfig -> (ActionState -> IO r) -> IO r`.
1. Provide a way to wrap the application constructed with the above in something that mocks certain responses (we need to pretend a3 is running).
1. Change this test to use that instead of the hspec-wai machinery (which has no way to providing extra context on top of the `Application`, as far as I know).

@ChristianSi: I think you wrote this test.  I have questions:

1. Does the above plan make sense to you?
1. In the current test: isn't `with setupBackend` and `withApp smartA3Backend` redundant?

ChristianSi:
@fhartwig:

> In a discussion with @fisx we talked about only storing the path component of a user-path (i.e. not the entire URL) and construct the user resource path dynamically. I don't think this is a good idea anymore, since a3 also stores the user path on user activation, and we probably don't want Thentos' and a3's view of the user path to diverge.

I'm pretty sure that's not true. Internally, the A3 backend doesn't store paths at all, but Python objects. There is a "principals" object with a "users" object under it, and one object for each user under the "users" object. Paths for these objects are generated dynamically whenever A3 sends a response. We don't have to mimic the object hierarchy, but just storing local paths (starting with "/principals") sounds reasonable to me. Then we we don't have to modify the DB whenever the exposed URL of our backend changes.

ChristianSi:
@fisx:

> Does the above plan make sense to you?

No, I don't understand the plan and I don't know why the test fails. The A3 backend and, likewise, our backend when mimicking A3, should always send complete URLs, so the correct approach should be to make sure that that still happens.

ChristianSi:
@fisx:

> In the current test: isn't with setupBackend and withApp smartA3Backend redundant?

I don't know why it should be redundant since we need `withApp smartA3Backend` to mimic A3. The one thing that has changed is that is doesn't really have to be "smart" anymore (sending different replies based on the request URL), since we now send only one request to A3. So we could replace `withApp smartA3Backend` by `withA3fake`, but that's a purely cosmetic change.

fisx:
To clarify (I hope it does clarify anything) -

- I was confused by the two backends, but of course we need one for the proxy and one for the application server (a3).  Please disregard my confused questions.
- There are two options, we can access the default config value from `Thentos.Test.Config`, or we can access the default config value that lives in the `ActionState` in the `Application`.  Whenever we use hspec-wai, we can only access the `Application`, but not the `ActionState`, but we can always regress to the first option (access default config value).

ChristianSi:
We don't need this, as the actual problem was elsewhere.

There are some nice fixups in here, but I'll salvage them into another PR.

PR #414, opened 2015-12-09
I18n for purescript

fisx:
- script for refreshing translation tables.
- renamed Mula.purs to I18n.purs
- new api for translation
- better documentation


refresh-i18n/src/Main.hs
```diff
@@ -0,0 +1,110 @@
+{-# LANGUAGE BangPatterns        #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE ViewPatterns        #-}
```
fhartwig: This program doesn't compile. It's missing `OverloadedStrings`

fhartwig:
I've looked at these changes and don't insist on backing them out.

refresh-i18n/src/Main.hs
```diff
@@ -0,0 +1,110 @@
...
+
+
+i18nModule :: FilePath
+i18nModule = "I18n"
+
+main :: IO ()
+main = do
+    setCurrentDirectoryToTarget
+    purss :: [FilePath] <- getTranslateableFiles "."
+    transKeys :: [ST] <- nub . sort . foldl' union [] <$> mapM translationKeys purss
```
fhartwig: Unless I'm missing something, this is *very* inefficient. `union` has at least quadratic complexity, so the whole thing is O(n^3) or worse. If this becomes a problem, it should be easy to fix by using `Data.Set`

PR #415, opened 2015-12-12
Binary python package

fisx:
Python egg constructor for thentos-adhocracy.

This is providing the compiled executable and config file as data files to python applications, with a small wrapper script.  no platform sensitivity yet.


ChristianSi:
Fine.

PR #416, opened 2015-12-12
Purescript delivery test case.

fisx:
This is a minor issue, but it has been asked where the 'honour-it.js' file is coming from, and this patch will make its source go away.  (If you have one in your working copy, please delete by hand once.)


ChristianSi:
Looks good.

PR #417, opened 2015-12-15
Fix URL rendering

ChristianSi:
The generated paths were wrong if the default port is used (443 for https, 80 for http).


fisx:
i made way too many comments.  (-:

looks good otherwise, please decide yourself what you want to ignore and if you want to merge.

PR #418, opened 2015-12-15
Record all signup attempts in DB

fhartwig:
This is a sort of work-in-progress on https://tree.taiga.io/project/fisx-thentos/task/44
It could probably need some more tests and I'm still working on CSV output for the new api endpoint, but apart from that I think it's ready for review.


thentos-core/src/Thentos/Transaction.hs
```diff
@@ -618,6 +618,16 @@ garbageCollectCaptchas timeout = void $ execT
...
+-- | store signup attempt in the db
+recordSignupAttempt :: UserName -> Bool -> ThentosQuery e ()
+recordSignupAttempt name captchaCorrect = void $ execT
+    [sql| INSERT INTO signup_attempts (user_name, captcha_correct, timestamp)
+           VALUES (?, ?, now()) |] (name, captchaCorrect)
+
+getAllSignupAttempts :: ThentosQuery e [SignupAttempt]
+getAllSignupAttempts = map (\(n, cc, t) -> SignupAttempt n cc t) <$>
+    queryT [sql| SELECT user_name, captcha_correct, timestamp FROM signup_attempts |] ()
+
```
fisx: missing empty line.

thentos-tests/tests/Thentos/ActionSpec.hs
```diff
@@ -112,6 +112,23 @@ spec_user = describe "user" $ do
...
+    describe "addUnconfirmedUserWithCaptcha" $ do
+        it "records signup attempts in the database" $ \sta -> do
+            let ActionState (conns, _, _) = sta
+                name = UserName "someName"
+                userData = UserFormData name "secret" (forceUserEmail "me@example.com")
+                cid = "myCaptchaId"
+                solution = "theSolution"
+                captchaSolution = CaptchaSolution cid solution
+                req = UserCreationRequest userData captchaSolution
+            Right () <- runVoidedQuery conns $ T.storeCaptcha cid solution
```
fisx: you don't need to store the captcha solution: validity is not a requirement for this test.

thentos-tests/tests/Thentos/ActionSpec.hs
```diff
@@ -112,6 +112,23 @@ spec_user = describe "user" $ do
...
+            void . runPrivs [RoleAdmin] sta $ addUnconfirmedUserWithCaptcha req
```
fisx: This has to work without `RoleAdmin`.

thentos-tests/tests/Thentos/ActionSpec.hs
```diff
@@ -112,6 +112,23 @@ spec_user = describe "user" $ do
...
+            [(name', captchaCorrect)] <- doQuery conns
+                [sql| SELECT user_name, captcha_correct
+                      FROM user_add_attempts |] ()
+            name `shouldBe` name'
+            captchaCorrect `shouldBe` True
```
fisx: as above: different test?

fisx:
otherwise looks good!

ChristianSi:
Though I notice it's a requirement, I'm very much against adding a publicly accessible endpoint (`registration_attempts`) revealing such sensitive information. Alternatives:

* Don't make this part of the API at all, instead write a little Perl script that reads the DB table and converts it into csv format.
* If part of the API, make sure that the endpoint is only accessible from certain configurable IPs (default: localhost).

thentos-tests/tests/Thentos/ActionSpec.hs
fhartwig: I meant to write a test for each case (with and without valid captcha) to make sure we store the attempt either way.

fisx:
as agreed offline: let's go for the privileged-ip solution.

thentos-tests/tests/Thentos/ActionSpec.hs
fisx: > I meant to write a test for each case (with and without valid captcha) to make sure we store the attempt either way.

ok, that makes sense.

fhartwig:
I didn't get around to doing the IP authentication. I'm unassigning myself as I'll be away for the next 2 weeks. Apart from the authentication issue, I think this is good to review/merge.

thentos-core/src/Data/Csv/Missing.hs
```diff
@@ -0,0 +1,17 @@
...
+import Control.Monad (mzero)
+
+instance ToField Bool where
+    toField True = "1"
+    toField False = "0"
+
+instance FromField Bool where
+    parseField "1" = pure True
+    parseField "0" = pure False
+    parseField _   = mzero
```
fisx: See also: https://github.com/hvr/cassava/pull/107

thentos-core/src/Thentos/Backend/Api/Simple.hs
```diff
@@ -87,6 +88,7 @@ type ThentosUser =
   :<|> "activate" :> ReqBody '[JSON] (JsonTop ConfirmationToken)
                   :> Post '[JSON] (JsonTop ThentosSessionToken)
   :<|> "login" :> ReqBody '[JSON] LoginFormData :> Post '[JSON] (JsonTop ThentosSessionToken)
+  :<|> "registration_attempts" :> Get '[JSON, (CSV', DefaultEncodeOpts)] [SignupAttempt]
```
fisx: Why is the path so different from the name of the handler function?

fisx:
What about using hs-logger for this?  We could define a dedicated variant of logger that takes all the information on the event (signup attempt in this case), stores it to a dedicated `Logger` (which goes to a separate file), and then do all the processing of this file outside of the `thentos` server process?

Benefits:

- @fhartwig, I suspect you might like that because it obsoletes the idea of using `PrivilegedIP` for authorization in the retrieval action (see #420).
- It would also be a lot more extensible: If we come accross some pattern of requests (either for thentos or for adhocracy) that we want to keep an eye on, we can easily and rapidly add them.
- It would also be nice to be able to patch the analytics code independently of the production server.

fhartwig:
Closing this in favour of https://github.com/liqd/thentos/pull/445

PR #420, opened 2015-12-16
Privileged IPs

fisx:


ChristianSi:
Looks good! The privileged IPs should, of course, be read from the config instead of being hardwired in the code, but otherwise this is exactly what I had in mind.

fhartwig:
No objections.

PR #421, opened 2015-12-16
Proposal for the messaging API between Thentos and A3 (work in progress)

ChristianSi:


docs/messaging.md
```diff
@@ -0,0 +1,58 @@
...
+* html: the HTML-formatted body of the message
+
+The following fields are alternative--exactly one of them must be present:
+
+* target_users: a list of one or more strings where each string is the user
+  path of a user. The message will be send to each of these users.
+  Implementation note: *users* in a service (such as A3) are *personas* in
+  Thentos. In Thentos, a service's notation of a user path is stored as
+  *ExternalUrl* of the persona.
+* target_email: a list of one or more strings where each string is an email
```
fisx: `target_emails`?  or singular in both cases and accept both a single value and a list as argument?  i prefer the latter.

docs/messaging.md
```diff
@@ -0,0 +1,58 @@
...
+  address. The message will be send to each of these addresses.
+
+## Messaging Reply
+
+If Thentos could successfully send all emails, it replies with status
+200 OK and the following JSON payload:
+
+    {"data": {}}
+
+FIXME Or maybe just send an empty object (`{}`) instead?
```
fisx: or empty body, no content_type?  not sure how that works in rest/http.

docs/messaging.md
```diff
@@ -0,0 +1,58 @@
...
+
+If the request was malformed, it replies with status 400 Bad Request and an
+error description in JSON format. FIXME Document the typical/expected cases.
+
+Services need to run on a privileged IP in order to be able to send
+messages. By default, only localhost is privileged, but that can be changed
+by modifying the Thentos config file (FIXME document config setting). If
+somebody on a privileged IP address send a POST request to `message`
```
fisx: "... send*s*..."

fisx: "... on a *non*-privileged..."?

docs/messaging.md
```diff
@@ -0,0 +1,58 @@
...
+endpoint, Thentos replies with 401 Unauthorized.
+
+If the messages could not be send due to some internal problem not caused
+by the sender, Thentos replies with 500 Internal Server Error.
```
fisx: if the error lies beyond thentos in the smtp world, i'm not sure 500 is the right one.  perhaps then it depends on the smtp error?

fisx:
Thanks, this is a good starting point.

I've just been thinking that it would also be nice to have `servant-python` and supply a client library to make it easier to use from a3.  Just an idea.

docs/messaging.md
```diff
@@ -0,0 +1,58 @@
...
+* body: the plain-text body of the message
+
+The following field is optional:
+
+* html: the HTML-formatted body of the message
+
+The following fields are alternative--exactly one of them must be present:
+
+* target_users: a list of one or more strings where each string is the user
+  path of a user. The message will be send to each of these users.
```
fhartwig: "sent", not "send"

docs/messaging.md
```diff
@@ -0,0 +1,58 @@
...
+  Implementation note: *users* in a service (such as A3) are *personas* in
+  Thentos. In Thentos, a service's notation of a user path is stored as
+  *ExternalUrl* of the persona.
+* target_email: a list of one or more strings where each string is an email
+  address. The message will be send to each of these addresses.
```
fhartwig: "sent"

PR #423, opened 2015-12-21
Shrink dependency footprint.

fisx:
After this goes to master, after the next update, all previously cloned repos may need to do this:

```
rm -rf .git/modules/submodules/servant-purescript
rm -rf .git/modules/submodules/servant
cabal sandbox delete-source .../submodules/servant/servant-session
./misc/thentos-install.sh
```


ChristianSi:
Looks OK (I made one line note).

PR #424, opened 2015-12-22
Avoid some ScopedTypeVariables

thielema:


fisx:
Interesting, thanks!  I like some of the changes (`\(SomeException _) -> ...`), but some things looked more concise and elegant to me before (`do x :: T <- ...`).

Is there a technical reason not to turn on ScopedTypeVariables by default in all modules, whether we need it or not?  Or do you just generally opose non-Haskell98 features?

thielema:
On Tue, 22 Dec 2015, fisx wrote:

> Interesting, thanks! I like some of the changes (\(SomeException _) -> 
> ...), but some things looked more concise and elegant to me before (do x 
> :: T <- ...).

You can just move the (:: T) type annotation wherever you want and it will 
not cost you extra characters. However, I think that type signatures are 
more principled than type annotations. E.g. GHC will warn about a missing 
type signature if you write (x = expr :: T) at top-level.

> Is there a technical reason not to turn on ScopedTypeVariables by 
> default in all modules, whether we need it or not? Or do you just 
> generally opose non-Haskell98 features?

I like to show that you can avoid some extensions with only little effort.

thielema:
I have replaced some more ScopedTypeVariables by Proxy combinators.

fisx:
> I like to show that you can avoid some extensions with only little effort.

But that's my point.  Of course every program can be written in Haskell98.   Language extensions have been introduced to make the language more expressive, so if you propose avoiding use of an extension, you have to give a reason other than it can be done.

fisx:
...  and I don't think the warning for missing top-level type signatures makes a good case for requiring where- and let-binding signatures.  That would mean that *all* local bindings must have signatures, not just the ones with annotations.

thielema:
On Tue, 22 Dec 2015, fisx wrote:

>       I like to show that you can avoid some extensions with only little effort.
> 
> But that's my point. Of course every program can be written in 
> Haskell98. Language extensions have been introduced to make the language 
> more expressive, so if you propose avoiding use of an extension, you 
> have to give a reason other than it can be done.

I see the burden of proof the other way round: If you want to use an 
extension it should solve a real problem. Saving a few characters is not 
enough justification for me. Every extension reduces portability to other 
compilers, GHC versions, operating systems. Some features have flaws in 
some GHC versions or their behaviour changes slightly. The less extensions 
you depend on, the better.

thielema:
On Tue, 22 Dec 2015, fisx wrote:

> ... and I don't think the warning for missing top-level type signatures 
> makes a good case for requiring where- and let-binding signatures.

I have maintained signatures in some cases in case someone added them for 
documentation purposes.

> That would mean that all local bindings must have signatures, not just 
> the ones with annotations.

In some cases I removed local signatures completely because they were not 
necessary for the compiler. But maybe this was wrong because a human 
wanted to read them?

So generally I have no opinion whether you should add or omit local type 
signatures, but I found that some type annotations were added in a way 
that required ScopeTypeVariables - without need. You can either omit them 
or turn them into local type signatures or move them from the binding to a 
use site. There is no problem to solve using ScopeTypeVariables.

thielema:
On Wed, 23 Dec 2015, thielema wrote:

> So generally I have no opinion whether you should add or omit local type 
> signatures, but I found that some type annotations were added in a way that 
> required ScopeTypeVariables - without need. You can either omit them or turn 
> them into local type signatures or move them from the binding to a use site. 
> There is no problem to solve using ScopeTypeVariables.

If you don't miss ScopeTypeVariables too much you may tell me the priority 
of the three variants and I try to bring the affected code parts to your 
prefered form.

I may also prepare a patch containing only proxy-related changes. I think 
that it is more appropriate to connect proxies with functions like 
subProxy, instead throwing away old proxies and creating new (somehow 
unrelated) ones. My overall impression of Servant is, that it requires too 
much type gymnastics that is unrelated to the value level. I assume that 
this is the reason why they need all these Proxys. But I do not plan to 
rewrite Servant so for now this is out of scope for my proposals to 
Thentos.

fisx:
I guess you are right about compiler bugs in general, but I'm pretty confident that ScopedTypeVariables is as mature as core optimization that you are still relying on with Haskell98.  But this:

> Every extension reduces portability to other compilers, [...]

this is a non-argument.  There are other Haskell98-compilers, but none that I know of that compiles `base-4`.

> But maybe this was wrong because a human wanted to read them?

Yes, I am a slow thinker and need lots of types annotations / signatures to feel comfortably oriented.  (-:

> If you don't miss ScopeTypeVariables too much you may tell me the priority of the three variants and I try to bring the affected code parts to your prefered form.

I don't know if I have a preferred form, and I like the freedom of adding types in different places.  I can take your this branch and cherry-pick the changes I like, would that be ok?

> I may also prepare a patch containing only proxy-related changes. I think that it is more appropriate to connect proxies with functions like subProxy, instead throwing away old proxies and creating new (somehow unrelated) ones.

Interesting.  I'm not very happy with the way we use proxy types, mostly because they are so verbose. Perhaps you can start a branch with an example of what you mean?  Or add a comment with pseudo-code, whichever is easier for you.

Thanks for your thoughts, I really appreciate it!  Sorry for being so difficult.  (-:

thielema:
On Thu, 24 Dec 2015, fisx wrote:

>       I may also prepare a patch containing only proxy-related changes. I think that it is more appropriate to
>       connect proxies with functions like subProxy, instead throwing away old proxies and creating new (somehow
>       unrelated) ones.
> 
> Interesting. I'm not very happy with the way we use proxy types, mostly 
> because they are so verbose. Perhaps you can start a branch with an 
> example of what you mean?

I refered to parts of the changes in the pull-request regarding 
ScopedTypeVariables. You may grep for subProxy.

fisx:
@thielema: I have cherry-picked your changes, see #437.  The proxy trick looks interesting, but I feel that keeping the way we do it now is both more common in the servant community and easier to read (if a bit more awkward to write).

I will merge the changes as soon as travis goes green.  Thanks!

thielema:
On Thu, 24 Dec 2015, fisx wrote:

> I guess you are right about compiler bugs in general, but I'm pretty 
> confident that ScopedTypeVariables is as mature as core optimization 
> that you are still relying on with Haskell98.

https://ghc.haskell.org/trac/ghc/ticket/11339

fisx:
> > I guess you are right about compiler bugs in general, but I'm pretty
> > confident that ScopedTypeVariables is as mature as core optimization
> > that you are still relying on with Haskell98.
>
> https://ghc.haskell.org/trac/ghc/ticket/11339

Oops.  :)

It looks like this has more to do with RankNTypes than with
ScopedTypeVariables.  Also, failure to produce an expected proof is
fairly benign since it can't lead to run-time or even security
problems.

But you certainly made your point.

Let's see what the others think when we're all back at work tomorrow.

thielema:
On Sun, 3 Jan 2016, fisx wrote:

> > > I guess you are right about compiler bugs in general, but I'm pretty
> > > confident that ScopedTypeVariables is as mature as core optimization
> > > that you are still relying on with Haskell98.
> >
> > https://ghc.haskell.org/trac/ghc/ticket/11339
> 
> Oops. :)

To be honest, you can also run into problems with plain Haskell 98 types:
    https://ghc.haskell.org/trac/ghc/ticket/11397

In this case I found a work-around using ScopedTypeVariables.

thielema:
On Sun, 10 Jan 2016, thielema wrote:

> On Sun, 3 Jan 2016, fisx wrote:
>
>> > > I guess you are right about compiler bugs in general, but I'm pretty
>> > > confident that ScopedTypeVariables is as mature as core optimization
>> > > that you are still relying on with Haskell98.
>> >
>> > https://ghc.haskell.org/trac/ghc/ticket/11339
>> 
>> Oops. :)
>
> To be honest, you can also run into problems with plain Haskell 98 types:
>   https://ghc.haskell.org/trac/ghc/ticket/11397
>
> In this case I found a work-around using ScopedTypeVariables.

But then again, since the bug breaks Haskell 98 compatibility it got 
highest priority.

PR #425, opened 2015-12-22
avoid some 'return ()' by Foldable functions

thielema:


fisx:
Yes, much better!  Thanks :+1:

PR #426, opened 2015-12-22
replace some ViewPatterns by regular function applications

thielema:


fisx:
Sorry, but I like `ViewPatterns`.  I think it's good to avoid extra let bindings, extra lambdas, and extra lines of code.  I added a comment to the one change I can completely agree with, but I do not see any reason to take the other changes.

Any points I missed?

fisx:
I'm closing this in favour of #430.  If I have missed anything important, please re-open.  Thanks!

PR #427, opened 2015-12-22
Action, Action.Core: avoid TupleSection

thielema:


fisx:
:+1: thanks!

I had some comments, but overall this is an clear improvement, so I will merge it.

PR #428, opened 2015-12-22
Transaction.unregisterPersonaFromContext: avoid LambdaCase using Foldable.mapM_

thielema:


thentos-core/src/Thentos/Transaction.hs
```diff
@@ -322,10 +321,9 @@ registerPersonaWithContext persona sid cname = do
 -- context.
 unregisterPersonaFromContext :: PersonaId -> ServiceId -> ContextName -> ThentosQuery e ()
 unregisterPersonaFromContext persId sid cname = findContextId sid cname >>=
-    \case Just cxtId -> void $ execT
+    mapM_ (\cxtId -> void $ execT
                 [sql| DELETE FROM personas_per_context WHERE persona_id = ? AND context_id = ? |]
-                (persId, cxtId)
-          Nothing    -> pure ()
+                (persId, cxtId))
```
fisx: I like `LambdaCase`, but I also like your change, for it avoids handling the `Nothing` case explicitly.  Thanks!

PR #429, opened 2015-12-22
Transaction.checkUnique: common way to check SQL query results

thielema:


fisx:
No argument here :-).  Thanks!

PR #433, opened 2015-12-27
Rewrite ./misc/thentos-install.sh in Haskell.

fisx:
I made a stupid mistake when I was changing the semantics of the shell script, so I rewrote it to see how easy that is.  The new Haskell script is 1/3 longer than the shell script in terms of LOC.  No libraries besides base are used, and the overhead is mostly due to primitive command line parser, path detection etc.

This should make it easier to:
- change the script without making stupid mistakes
- add features like building purescript and haskell in parallel, or running all test suites in parallel


fhartwig:
Looks reasonable.

PR #438, opened 2016-01-03
bcrypt (work in progress)

fisx:
adhocracy3 stores user data with bcrypt-protected passwords.  for a seemless migration to thentos in production, we need to be able to re-use these passwords after they have been copied to the thentos database.  this PR explores how this could be done.


fisx:
I think the last commit would be considerably more concise if you would keep storing serialized `HashedSecret` values as we did before.  It would only require a parser and a renderer for `HashedSecret` (cereal?  binary?  i think we already use cereal somewhere?), and a new case in `secretMatches`.

Am I missing something here?

fisx:
closing in favour of #457

PR #440, opened 2016-01-04
documentation fix in README; setting PGUSER was not working when crea…

ssadler:
…ting database user


fisx:
like!  will wait for travis and merge.

PR #443, opened 2016-01-06
Arguments to Crypto.BCrypt.validatePassword where flipped

np:
Additional comments:

* For ASCII password no further encoding is necessary, having additional test cases with non-ASCII would help going further.
* Make sure to depend on memory>=0.10 (https://github.com/A1kmm/hs-bcrypt/issues/8 https://github.com/vincenthz/hs-memory/commit/17af888d3a34c2322a17b8556f45150c16a98fde)


ChristianSi:
Great, thanks!

PR #448, opened 2016-01-13
Path handling (across different dev and op scenarios)

fisx:
The original purpose of this branch is to get rid of absolute path names in compiled executables, which I suspect are served by cabal's `getDataFileName` in the generated module `Path_package_name`.

The config type now allows for a field `root_path` that is `.` by default.  Since this is configifier, you can also use the cli or set `ROOT_PATH` in the shell environment if `.` is not adequate.

I also cleaned up the test suite to work with the new content type and create databases and action states in a more coherent way.

### DIFF
--- a/.travis/script.sh
+++ b/.travis/script.sh
@@ -19,4 +19,4 @@ cp -r travis-cache/data ./hlint/.cabal-sandbox/share/x86_64-linux-ghc-7.8.4/hlin
 sed -i 's/^jobs:/-- jobs:/' ${HOME}/.cabal/config
 
 cabal update
-./misc/thentos-install.hs -c "--force-reinstalls" -t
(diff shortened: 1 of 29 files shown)


fisx:
(travis fails because `ppa:hvr` is unresponsive.  just for an hour or so?)

ChristianSi:
Guess my comments points to an issue that needs to be addressed (though I cannot currently see why it doesn't work), otherwise this looks good.

fisx:
(doing some heavy re-basing again, sorry...)

fisx:
There are still a few errors in the test suite and I would like to update the documentation when those are fixed, but other than that this PR is ready for a preliminary review.

ChristianSi:
Looks good, please merge after warnings have been fixed.

PR #450, opened 2016-01-13
Fix audio captchas and generally discard whitespace in captchas

ChristianSi:
Audio captchas where stored in the form "1 2 3 4 5 6", so when users typed "123456", the solution was rejected. Now whitespace is generally ignored and discarded, for both visual and audio captchas. This also means that users can submit visual captcha solutions in the form "abc def" or "abcdef", and both will be accepted.


fisx:
Looks good!  I'll wait for travis and merge.

fisx:
The purescript problem travis has has nothing to do with this branch.  I'll merge anyway.

PR #451, opened 2016-01-14
Fix random digit generation for audio captcha

fhartwig:
The previous code would generate '3' for every input byte > 127, so more
than half of the generated digits would be '3'.


fhartwig:
(Note: the digits generated are still somewhat biased since 10 doesn't divide 256, but this should be good enough for this purpose)

fisx:
Never mind.  I agree this is good enough.  Will wait for travis and merge.

fhartwig:
Fixed the obsolete import.

fisx:
purescript is wobbly today.  since this is not related to this PR i will merge.

PR #457, opened 2016-01-20
Support import of bcrypt data.

fisx:
Support importing users with `bcrypt`-protected passwords.

- Store `HashedSecret` as `bytea` in DB.
- Change names of `verify*` functions.
- Refactor `hashSecret`, `hashSecretWith`: separate pure code from effects.
- Re-think test case in `ActionSpec`.
- Tests.


ChristianSi:
Looks good (besides the few comments I made).

fisx:
Non-trivial changes; could you take another look?

ChristianSi:
Looks good.

PR #458, opened 2016-01-21
Script for importing a3 users

fisx:


ChristianSi:
Looks good. I made some comments, none of them essential, so please merge after addressing them or not (if Travis agrees).

PR #461, opened 2016-01-21
Integrate new password reset into thentos-a3 api

### LINKED ISSUE #321: Change password reset so that Thentos handles it without requiring A3 in any way
Backend + frontend

### LINKED ISSUE #407: Don't export _changePasswordUnconditionally function
The `_changePasswordUnconditionally` function shouldn't be exported, but wrapped in a public action that establishes that the password change is legitimate. See issue #321.

See BUG marker in `thentos-core/src/Thentos/Action.hs`.

ChristianSi:
This largely fulfills Taiga Task 75: Integrate password reset endpoints into thentos-adhocracy API.

UPDATE: this PR should now be complete and has been successfully tested to work with A3 in proxy mode.

Fixes #321 (the A3 backend is no longer needed; the A3 frontend is still used, but that's intentional). Also fixes #407.

~~Also, I disabled a test that checks whether short passwords are rejected -- in A3 they are, in Thentos not yet. It makes almost certainly sense to enforce a lower limit, but that also affects all other actions where passwords are set (create user, update password etc.) and it thus unrelated to this PR.  (see #463)~~

### DIFF
--- a/thentos-adhocracy/devel.config
+++ b/thentos-adhocracy/devel.config
@@ -14,6 +14,7 @@
 # * Call bin/buildout
 
 command: "run"
+root_path: "../thentos-core"
 
 backend:
     bind_port: 6546
@@ -81,3 +82,14 @@ email_templates:
             This is a reminder that you already have a Thentos account. If you
             haven't tried to sign up to Thentos, you can just ignore this email.
             If you have, you are hereby reminded that you already have an account.
+    password_reset:
+        subject: "Thentos: Reset Password"
+        # Supported variables: {{user_name}}, {{reset_url}}
+        body: |
+            Dear {{user_name}},
+
+            please use the link below to reset your password.
+
+            {{reset_url}}
+
+            Your Thentos Team
--- a/thentos-adhocracy/src/Thentos/Adhocracy3/Action.hs
+++ b/thentos-adhocracy/src/Thentos/Adhocracy3/Action.hs
@@ -11,17 +11,14 @@ module Thentos.Adhocracy3.Action
     , login
     , makeExternalUrl
     , resetPassword
-    , userIdFromPath
     ) where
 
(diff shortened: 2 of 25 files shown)


fisx:
> (note the additional `?path=` in the URL.) This can be easily fixed by passing a `beforeToken :: Maybe String` argument to `addPasswordResetToken`, but it's not yet done.

I don't follow.  We are only responsible for the backend, and there we currently have (in thentos-adhocracy):

```haskell
  :<|> "password_reset"        :> ReqBody '[JSON] PasswordResetRequest
                               :> Post200 '[JSON] RequestResult
```

isn't it enough to simply change that to this?:

```haskell
  :<|> "password_reset"        :> Capture "token" PasswordResetToken
                               :> Post200 '[JSON] RequestResult
```

fisx:
Looks good so far!

ChristianSi:
@fisx re `?path=` fragment: this fragment must appear in an URL handled by the A3 frontend, so it's unrelated to Servant. We have to modify the email-sending function to generate links in the form expected by the A3 frontend.

ChristianSi:
The following changes have been implemented by now: Our ~~`addPasswordResetToken`~~`sendPasswordResetMail` action (`create_password_reset` endpoint) generates frontend paths of the form:

     http://frontend_url/password_reset/TOKEN

where `TOKEN` is the reset token passed on to the backend. But in A3, the URLs have the form

     http://frontend_url/password_reset/?path=TOKEN

(note the additional `?path=` in the URL.) This can be easily fixed by passing a `beforeToken :: Maybe String` argument to `sendPasswordResetMail`, but it's not yet done.

Another small issue missing: copy email templates from `thentos-core/devel.config` to `thentos-adhocracy/devel.config`.

fisx:
LGTM.  will wait for travis and merge.

Issue #455, opened 2016-01-18
allow_ips (privileged-IP check) doesn't support IPv6


ChristianSi:
The thentos config has a "allow_ips" field that can be used to list privileged IP addresses that are authorized to perform some actions not available from other IP addresses. Currently, allowedIps is a `[HostAddress]`, that is, only IPv4 addresses can be listed.

To make this generally usable, we should instead define and use a
sum type that accepts both `HostAddress` and `HostAddress6`. Then we can resolve the FIXME  in `thentos/thentos-core/src/Thentos/Backend/Core.hs` to related to `SockAddrInet6` by checking whether that address is privileged.

fisx:
Fixed in #469.

(@np: if you add the string: `Fixes #<issue number>` into the description of a PR, merging will auto-close the issue.)

PR #464, opened 2016-01-25
[Issue #455] Add support for IPv6 addresses in allow_ips configuration

np:
* Instead of `inet_addr`, use `getAddrInfo` which supports IPv6
* A new module `Network.HostAddr` including:
    * `HostAddr` a type wrapping the different kind of addresses
    * `getHostAddr` a wrapper over `getAddrInfo` returning a `HostAddr`
    * `hostAddr` extract `HostAddr` from a `SockAddr`

### DIFF
--- a/thentos-core/src/Network/HostAddr.hs
+++ b/thentos-core/src/Network/HostAddr.hs
@@ -1,0 +1,35 @@
+{-# LANGUAGE LambdaCase #-}
+
+module Network.HostAddr
+where
+
+import Control.Monad.Except (throwError)
+import Data.Int (Int32)
+import Network.Socket
+          (SockAddr(SockAddrInet, SockAddrInet6, SockAddrUnix, SockAddrCan),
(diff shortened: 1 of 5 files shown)


fisx:
Nice, thanks!

I just realized that https://github.com/liqd/thentos/blob/master/docs/styleguide.md#layout does not mention that we indent 4 spaces, not two.  Now I'm wondering if we should add that.  @np, would you have strong objections?

fisx:
(you can simulate travis checks locally by running `./misc/thentos-intsall.hs -p -t`.  see `-h` for a one-line explanation.  of course it's also ok to have travis do the work.  :-)

fisx:
This looks pretty much done, just a few small questions.  Please unassign yourself if this is ready for review.

Thanks!

np:
I made another pull request since both the date was wrong and the branch was in my repo.

PR #468, opened 2016-01-25
Add a stack.yml file

np:
These are not used directly but were added to please Cabal at some point:
  asn1-encoding ==0.9.2
  asn1-parse ==0.9.1
  asn1-types ==0.3.0
  x509 ==1.6.1
  x509-validation ==1.6.2

The Data.EitherR.fmapL is used from the errors package. This relax its
constraints to (>=2.0.0 && <2.1).


fisx:
There was some interesting and non-trivial issue that made us add these constraints, but I don't remember the details and both travis and my laptop seem to be happy with removing them.  LGTM, then!

PR #469, opened 2016-01-25
[Issue #455] Add support for IPv6 addresses in allow_ips configuration

np:
* Instead of `inet_addr`, use `getAddrInfo` which supports IPv6
* A new module `Network.HostAddr` including:
    * `HostAddr` a type wrapping the different kind of addresses
    * `getHostAddr` a wrapper over `getAddrInfo` returning a `HostAddr`
    * `hostAddr` extract `HostAddr` from a `SockAddr`

### DIFF
--- a/thentos-core/src/Network/HostAddr.hs
+++ b/thentos-core/src/Network/HostAddr.hs
@@ -1,0 +1,33 @@
+{-# LANGUAGE LambdaCase #-}
+
+module Network.HostAddr
+where
+
+import Data.Int (Int32)
+import Network.Socket
+          (SockAddr(SockAddrInet, SockAddrInet6, SockAddrUnix, SockAddrCan),
+           HostAddress, HostAddress6, SocketType(Stream), AddrInfoFlag(AI_NUMERICHOST),
+           addrAddress, getAddrInfo, addrFlags, addrSocketType, defaultHints)
+
+data HostAddr
+    = HostAddress  HostAddress
+    | HostAddress6 HostAddress6
+    | UnixAddress  String
+    | CanAddress   Int32
+    deriving (Eq, Read, Show)
+
+hostAddr :: SockAddr -> HostAddr
+hostAddr = \case
+    SockAddrInet _ ip -> HostAddress ip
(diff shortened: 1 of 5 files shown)


thentos-core/src/Network/HostAddr.hs
```diff
@@ -0,0 +1,33 @@
...
+    SockAddrUnix fp -> UnixAddress fp
+    SockAddrCan i -> CanAddress i
+
+getHostAddr :: String -> IO HostAddr
+getHostAddr name = hostAddr . addrAddress . single <$> getAddrInfo (Just hints) (Just name) Nothing
+  where
+    hints = defaultHints { addrFlags = [AI_NUMERICHOST], addrSocketType = Stream }
+
+    single [x] = x
+    single []  = error "Impossible error: numeric addresses should always resolve"
```
fisx: [Your last rebase killed my earlier comments, so I'll add it to the "changed files" tab and not to the commit this time.]

"impossible error" means "programmer failed", but this is an instance of "configuration file is bad" (at least potentially).  I would prefer:

```haskell
single [] = throwIO $ ErrorCall "Configuration error: numeric address should always resolve"
```

or even change the result type to `Either String HostAddr` and deal with the error at the call site.

I prefer `throwIO` to `error` even if the exception type is no more specific, because it works without the implicit `unsafePerformIO`.  Does that make sense?

fisx:
Thanks @np, could you comment / act on my last remark and pass back to me?  If you don't want to change your error handling I guess we can argue that the `init*` function I mentioned earlier should rule out both errors at start time.

fisx:
(I just found #464 with the old comments.  Sorry for the noise!)

thentos-core/src/Network/HostAddr.hs
np: Actually, when the configuration ill formatted this is the `getAddrInfo` which throws an `IOException`. The test case might not be explicit enough but I intended these "Impossible error" as "programmer failed".

np:
I would wait for this `init*` function, in particular if we can't trigger these "impossible" cases.

fisx:
Makes sense now.  Sorry for being so slow.  :-)

Issue #465, opened 2016-01-25
Refactoring: _Left subsumes fmapL* functions


np:
These functions: `fmapL`, `fmapLM`, `fmapLTM` are subsumed by the use of the `_Left` prism from the lens library. Moreover this seems to be the only imports to the errors package.

fisx:
:+1:

(haven't checked, but sounds like I good change.)

fisx:
Fixed in #470

PR #470, opened 2016-01-25
[Issue #465] Refactoring: _Left subsumes fmapL* functions

np:
* Removes the dependency on `errors`
* `fmapL` can be replaced by `over _Left` or `first`
* `fmapLM` can be replaced by `_Left` or `traverseOf _Left`
* `fmapLTM` was no longer used


fisx:
LGTM.

Issue #472, opened 2016-01-26
Sending emails through Thentos


np:
A new API endpoint should allow services to send emails to users.

Sending an email takes the following fields:

* recipient (email address, pseudonym)
* subject
* plain text body
* optional HTML body

Where to put this endpoint? Under service/ at the root?

fisx:
> Where to put this endpoint? Under service/ at the root?

A service in thentos is a web app that uses thentos for identity management, like adhocracy.  Our data model anticipates many services for one instance, and `/service` is the sub-tree of the api that allows CRUDding them.

What about this `/email`?

np:
`/email` as a group and then `/email/send` or just `/email`?

fisx:
both good.  `/email/send` is more explicit, but i also like the approach of using http verbs to distinguish different things to do under one end point.

what else would there be besides `/email/send`?

np:
I'm now convinced that a `POST` request `/email` is enough.

fisx:
Fixed in #473.

PR #473, opened 2016-01-26
Basic support to send emails [#472]

np:
No support yet for the HTML optional body.

### DIFF
--- a/thentos-core/src/Thentos/Action.hs
+++ b/thentos-core/src/Thentos/Action.hs
@@ -246,7 +246,7 @@ sendUserConfirmationMail user (ConfirmationToken confToken) = do
         context "activation_url" = MuVariable $ exposeUrl feHttp <//> "/activate/" <//> confToken
         context x                = error $ "sendUserConfirmationMail: no such context: " ++ x
     body <- U.unsafeAction $ U.renderTextTemplate bodyTemplate (mkStrContext context)
-    U.unsafeAction $ U.sendMail (Just $ udName user) (udEmail user) subject (cs body)
+    U.unsafeAction $ U.sendMail (Just $ udName user) (udEmail user) subject (cs body) Nothing
 
 -- | Helper action: Send a confirmation mail to a newly registered user.
 sendUserExistsMail :: UserEmail -> Action e s ()
@@ -254,7 +254,7 @@ sendUserExistsMail email = do
     cfg <- U.unsafeAction U.getConfig
     let subject = cfg >>. (Proxy :: Proxy '["email_templates", "user_exists", "subject"])
         body    = cfg >>. (Proxy :: Proxy '["email_templates", "user_exists", "body"])
-    U.unsafeAction $ U.sendMail Nothing email subject body
+    U.unsafeAction $ U.sendMail Nothing email subject body Nothing
 
 -- | Initiate email-verified user creation.  Does not require any privileges, but the user must
 -- have correctly solved a captcha to prove that they are human.  After the new user has been
@@ -326,7 +326,7 @@ sendPasswordResetMail beforeToken email = do
         context "reset_url" = MuVariable $ exposeUrl feHttp <//> "/password_reset/" <//> fullTok
         context x           = error $ "sendPasswordResetMail: no such context: " ++ x
     body <- U.unsafeAction $ U.renderTextTemplate bodyTemplate (mkStrContext context)
-    U.unsafeAction $ U.sendMail (Just $ user ^. userName) (user ^. userEmail) subject (cs body)
+    U.unsafeAction $ U.sendMail (Just $ user ^. userName) (user ^. userEmail) subject (cs body) Nothing
 
 -- | Finish password reset with email confirmation. Also confirms the user's email address if
 -- that hasn't happened before. Returns the ID of the updated user.
@@ -396,7 +396,7 @@ requestUserEmailChange uid newEmail callbackUrlBuilder = do
 
     let message = "Please go to " <> callbackUrlBuilder tok <> " to confirm your change of email address."
         subject = "Thentos email address change"
-    U.unsafeAction $ U.sendMail Nothing newEmail subject message
+    U.unsafeAction $ U.sendMail Nothing newEmail subject message Nothing
 
 -- | Look up the given confirmation token and updates the user's email address iff 1) the token
 -- exists, 2) the token belongs to the user currently logged in, and 3) the token has not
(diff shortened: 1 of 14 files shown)


np:
This ready for a first round of reviews. I still have some issues with my setup, so I've not ran the tests myself.

thentos-core/src/Thentos/Backend/Api/Simple.hs
```diff
@@ -161,6 +163,11 @@ thentosServiceSession =
   :<|> endServiceSession
 
 
+-- * send email
+
+type SendEmail =
+    ReqBody '[JSON] SendEmailRequest :> Post '[JSON] ()
+
 -- * servant docs
```
fisx: There are some layout style rules for haddock sections:

https://github.com/liqd/thentos/blob/master/docs/styleguide.md#layout

(There is also the rule that if you don't like anything in the style guide, you can write a PR.  And regarding haddock section headings, I'm not sure they even make sense in the presence of explicit export list.  I think in that case they should go to the export list, so this rule needs some work.  Anyway, this is what it is now...)

thentos-core/src/Thentos/Types.hs
```diff
@@ -22,7 +22,7 @@ module Thentos.Types
     , PasswordResetToken(..)
     , PasswordResetRequest(..)
     , UserFormData(..)
-    , UserCreationRequest(..)
+    , UserCreationRequest(..), emailRecipient, emailSubject, emailPlainTextBody
```
fisx: Are these in the wrong line?

thentos-core/src/Thentos/Types.hs
```diff
@@ -507,6 +509,22 @@ instance ToJSON ByUserOrServiceId where
     toJSON (ByService i k) = Aeson.object [ "service" .= Aeson.toJSON (i, k)]
 
 
+-- * send email
+
+-- TODO: Service ID? Pseudonyms? From address?
```
fisx: For from address, you can use this from `Thentos.Config`:

```haskell
type SmtpConfig' =
      Maybe ("sender_name"    :> ST)  -- FIXME: use more specific type 'Network.Mail.Mime.Address'
  :*>       ("sender_address" :> ST)
...
```

fisx: Service ID is not required for now.

Pseudonyms: yes, that would be nice.  Do you need details on this?  Essentially, they are a services handle for a user, and need to be looked up, mapped to the user, and then lead to the email address.

thentos-core/src/Thentos/Types.hs
np: Yes, it's a mistake.

thentos-core/src/Thentos/Types.hs
np: I'm using the smtp config already, as I'm using the same function calls.

fisx: Ah, you were asking if we want the sender to be able to change that: no, default sender from the config is fine.

thentos-core/src/Thentos/Action.hs
```diff
@@ -613,6 +615,16 @@ getServiceSessionMetadata :: ServiceSessionToken -> Action e s ServiceSessionMet
 getServiceSessionMetadata tok = (^. srvSessMetadata) <$> lookupServiceSession tok
 
 
+-- * send emails
+
+-- | Send an email. Only a privileged IP is allowed to use this endpoint.
+sendEmail :: SendEmailRequest -> Action e s ()
+sendEmail req = do
+    void hasPrivilegedIP
```
fisx: I wonder if `hasPrivilegedIP` should be changed to make this `void` unnecessary?  (Can't be bothered to check myself, not necessarily part of this PR, just wondering...)

fisx:
Notes from the phone call:

- we want to be able to send to lists of recipients.
- "recipient" can be "persona", "email", "group".  how those are distinguished still needs to be designed.
- templating: like in the activate_user end point.
- mass mailing: not an issue yet.  we can trust privileged IPs.

afterthough: ./docs/messaging.md still contains a few fixmes, and i think most of them are things that can be designed using nothing but common sense and general knowledge.  @np, could you add that either here or in a separate PR?

ChristianSi:
> afterthough: ./docs/messaging.md still contains a few fixmes, and i think most of them are things that can be designed using nothing but common sense and general knowledge. @np, could you add that either here or in a separate PR?

There is one FIXME about the return value of the new endpoint. I proposed

    {"data": {}}

which, with some hindsight, doesn't make any sense. In your PR, the return value is `Post '[JSON] ()` which (I think) corresponds HTTP status code `204 No Content` and no body. I think this is very reasonable and the messaging.md should be adapted to document it.

fisx:
Clarification:

> - "recipient" can be "persona", "email", "group".  how those are distinguished still needs to be designed.

I meant `PersonaId` here.  This is available to the service and the default way to refer to a persona.

fisx:
Nice to have: send email to `ServiceGroup`.

There are `Group`s of users (organising things in thentos) and `ServiceGroup`s of personas (organising the same things on the service side, i.e. adhocracy).  See ./docs/terminology.md for more details.

The service may want to send email to all (transitive) members of a service group.  For this to work, I think there may be data types, transactions and actions missing involving the use (not creation and assignment) of service groups.  Currently, we have:

```sh
grep persona_groups thentos-core/schema/schema.sql thentos-core/src/Thentos/Transaction.hs
grep group_tree thentos-core/schema/schema.sql thentos-core/src/Thentos/Transaction.hs
```

Please ping me if you want me to elaborate on this.

fisx:
Updated earlier comment.  Now:

> I meant ~~`PersonaId`~~ `personaExternalUrl`

np:
Hum, thanks to git rebase the date of this commit should have been now and not yesterday.

np:
This is ready for another round of review @fisx @ChristianSi

fisx:
Looks good.  Just a few comments.

np:
While my last commit ec0071a might be better for incremental review, I'm going to rebase onto master to make it a proper PR.

np:
Actually I rebased but preserved the separate commits at least for now.

fisx:
thanks for keeping the individual commits!  losing them would have been annoying.

generally speaking, i was asked not to rebase in between incremental reviews, and now i see why :-).  i'll assume though that the old commits haven't changed and will just read the last one.

fisx:
I made two small commits.  Could you check if you have any objections and otherwise merge?

np:
LGTM

PR #476, opened 2016-01-26
Reject passwords with less than 6 chars

### LINKED ISSUE #463: Reject short passwords.
Affects all actions where passwords are set (reset password, create user, update password etc.).

I vote for having a constant

```haskell
minPasswordLength :: Int
```

in, say, `Thentos.Actions` or some other suitable location.  No need to complicate config file further.

ChristianSi:
Fixes #463.


fisx:
Added one more check; otherwise LGTM.  @ChristianSi, could you take a quick look?

Will merge if travis if you and travis ok it.

ChristianSi:
@fisx Thanks, I'd forgotten about that action because it's not part of the REST API. Merging now.

PR #480, opened 2016-01-27
Use `cat` instead of `/bin/cat` for sendmail_path

np:
Indeed the command `cat` might be elsewhere as in `NixOS` for instance.


fisx:
(I wonder if there is a way to document this better that we've both missed now?  Or perhaps this is another instance of an `init*` function we want that is called at start time to trigger errors early?)

np:
Like checking early on that the value of sendmail_path does exist?

fisx:
Yes.  Only that check happens implicitly in configifier, and I would like to add something that checks that it's actually a path to an existing file.

fisx:
(Anyway, later.  This should work for now.)

PR #481, opened 2016-01-27
Return all rest errors with content-type json

### LINKED ISSUE #474: Return all rest errors with content-type json.
We currently use `Thentos.Backend.Core.baseActionErrorToServantErr` (and a few related functions) to catch and render errors in `Thentos.Backend.Core.enter`, but parse errors in request bodies are thrown outside of the handler, so they are not rendered.

We need to consistently render all error responses with the above function, or something similar in spirit.

One solution would be to write a `Network.Wai.Middleware` and call it in `runWarpWithCfg`.  I can't think of another solution, and I think this one is reasonably clean.

ChristianSi:
Fixes #474. There is still a FIXME in the new code (not sure whether that matters in practice), but at least it covers the intended use case of converting aeson errors to JSON.


fisx:
Thanks!

PR #483, opened 2016-01-28
Add personasFromGroup to Thentos.Transaction

np:


fisx:
I think I found a real (non-whitespace-related :-) issue in your tests, otherwise LGTM.  Could you please check and object or merge?

Issue #482, opened 2016-01-28
Refactoring: Add a Thentos.Prelude module to capture some essential imports 


np:
I wonder how you feel about such a change?

For instance in Ling you can see https://github.com/np/ling/blob/master/Ling/Prelude.hs

fisx:
Personally I don't feel a strong need for it, but I can't think of any disadvantages (everybody is still free to ignore the Prelude module), so as long as at least one stake holder is in favor of it (which would be you), I'm in!

Do you want to write the PR?

fisx:
I also note that link follows the much more common convention of interleaving qualified and non-qualified imports.  Should we make either choice binding in the style guide?  Which one?

fisx:
I think this is done.  Please re-open if not.

PR #484, opened 2016-02-01
Refactoring: Add Thentos.Prelude to trim down the number of imports #482

np:


thentos-core/src/Thentos/Prelude.hs
```diff
@@ -0,0 +1,33 @@
+module Thentos.Prelude (module X) where
+
+import Control.Applicative as X
+import Control.Conditional as X hiding ((|>), (<|), (??))
+-- import Control.Exception.Lifted as X hiding (try)
```
fisx: what's this commented-out line?

fisx:
I'm a bit worried about implicit imports.  I guess it's fine if we make all imports implicit in Prelude, people can still grep for things they don't know there.  Would you be happy with that change?

Also, I wonder about name clashes.  I don't think we can do explicit re-exports, right?

Having that said, the import lists *do* look a lot tidier here.  I would like to get this to work somehow.

thentos-core/src/Thentos/Prelude.hs
np: Shouldn't be here.
I tried to make it part of Prelude but this would require to be a real Prelude, namely override the original one. Advantage: simpler to get rid of old version of functions such as this `try`, disadvantage: slightly longer to import this prelude as we have to either `import Prelude ()` or use `NoImplicitPrelude` option.

np:
Did you meant "all imports explicit in Prelude", at least its imported explicitly only once so I guess its fine by me. Still some modules such as Lens have a huge list of exports. My take would be to leave it implicit by default and make things explicit on a case by case basis.

If you wish I can make some kind of a best effort listing explicitly the import for most modules.

For name clashes, on a case by case basis, I would hide it from one of them when the hidden one is either unlikely to be used or less good that the other one.

thentos-core/src/Thentos/Prelude.hs
fisx: I'm ok with using `-XNoImplicitPrelude` globally.  If we can agree on the other open questions.

fisx:
> Did you meant "all imports explicit in Prelude", at least its imported explicitly only once so I guess its fine by me. Still some modules such as Lens have a huge list of exports. My take would be to leave it implicit by default and make things explicit on a case by case basis.

Yes, that's what I meant.  I agree that the import lists would be potentially huge, but we only need to import what we actually use.

> If you wish I can make some kind of a best effort listing explicitly the import for most modules.

If you think it's a stupid idea, we can leave the task to anybody who wants it to happen.  So I'm happy with either importing everything explicitly, or leaving it as is, but documenting in the module header that we welcome "explicitalizing" PRs.  (@andorp: do you have an opinion?  anybody else listening in?)

> For name clashes, on a case by case basis, I would hide it from one of them when the hidden one is either unlikely to be used or less good that the other one.

...  and then have the module that uses the prelude make a qualified import?  sounds good.

fisx:
There is still the commented-out line.  Most of the discussion after that leaves you free to not act on it, if I'm not forgetting anything.  Let me know if you think this can be merged.

PR #485, opened 2016-02-01
Refactoring

np:
* Move `finally` into the new module `Control.Monad.Except.Missing`
* Move `formH` into the new module `Servant.Missing`
* Add a `MonadAction` context synonym to be used instead of some
  occurrences of `Action`
* Use `for_` instead of `mapM_` when this reads better
* Use `$>` instead of `>>` and `return`
* Use `.=` instead of `modify` and `.~`
* Use `%=` instead of `modify` and `%~`
* Use `use` instead of `gets` and `^.`
* Use `preuse` instead of `gets` and a manual version of `^?`
* Factor the tab switching as `switchTab` and `switchTab'`

### DIFF
--- a/thentos-core/src/Control/Monad/Except/Missing.hs
+++ b/thentos-core/src/Control/Monad/Except/Missing.hs
@@ -1,0 +1,9 @@
+module Control.Monad.Except.Missing where
+
+import Control.Monad.Except (MonadError(catchError, throwError))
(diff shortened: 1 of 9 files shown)


fisx:
This is a considerable improvement, thanks!  I fixed a typo and eliminated a FIXME.  If you have no objections you can merge.

PR #487, opened 2016-02-02
Support aula user import

fhartwig:


thentos-core/exec/ImportAulaUsers.hs
```diff
@@ -0,0 +1,49 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module Main where
```
fisx: The file name suggests this should probably go to the aula repo.  But you add a FIXME in if you don't want to do it.

fhartwig: I can put it in aula. I mostly put it here because it was more convenient to develop that way.

PR #488, opened 2016-02-04
Renaming.

fisx:
`./docs/terminology.md` was not in sync with the code.  These changes fix wrong uses of the word `role`.


fisx:
@np, assigning this to you since you are the only one touching the code right now, and we talked about this earlier today.  could you please merge resp. reject?  thanks!

PR #490, opened 2016-02-05
CSRF protection

np:
Ready for review:
* docs
* code

TODO:
* tests

### DIFF
--- a/thentos-core/devel.config
+++ b/thentos-core/devel.config
@@ -1,3 +1,7 @@
+# This "secret" is hardcoded and public.
+# If you remove it, thentos will generate a new one for you.
+csrf_secret: 1daf3741e8a9ae1b39fd7e9cc7bab44ee31b6c3119ab5c3b05ac33cbb543289c
+
 backend:
     bind_port: 7001
     bind_host: "127.0.0.1"
--- a/thentos-core/src/Thentos.hs
+++ b/thentos-core/src/Thentos.hs
@@ -84,8 +84,12 @@ makeMain commandSwitch =
           , "I just securely generated such a secret for you:"
           , "  " <> show freshCsrfSecret
           , ""
-          , "You can either use the field csrf_secret field from the configuration, or the"
-          , "CSRF_SECRET environement variable, or the command line flag --CSRF_SECRET."
+          , "You can either add the following line to the configuration:"
+          , "csrf_secret: " <> show freshCsrfSecret
+          , "or set the following environement variable:"
+          , "THENTOS_CSRF_SECRET=" <> show freshCsrfSecret
+          , "or pass the follwing command line flag:"
+          , "--csrf-secret=" <> show freshCsrfSecret
           , "Beware of keeping this value private, hence you might not want to put it in"
           , "a configuration file which is version controlled."
           ]
--- a/thentos-core/src/Thentos.hs
+++ b/thentos-core/src/Thentos.hs
@@ -31,6 +31,7 @@ import Thentos.Action.Core (runActionWithPrivs)
 import Thentos.Action.Types (Action, ActionState(..), aStDb, aStConfig)
 import Thentos.Config
 import Thentos.Frontend (runFrontend)
+import Thentos.Frontend.CSRF (CsrfSecret(CsrfSecret), validFormatCsrfSecretField, genCsrfSecret)
 import Thentos.Smtp (checkSendmail)
 import Thentos.Transaction.Core (createDB, runThentosQuery)
 import Thentos.Types
@@ -71,6 +72,23 @@ makeMain commandSwitch =
 
         mFeConfig :: Maybe HttpConfig
         mFeConfig = Tagged <$> config >>. (Proxy :: Proxy '["frontend"])
+
+        csrfSecret :: Maybe ST
+        csrfSecret = config >>. (Proxy :: Proxy '["csrf_secret"])
+
+    when (isJust mFeConfig && not (validFormatCsrfSecretField csrfSecret)) $ do
+        CsrfSecret freshCsrfSecret <- genCsrfSecret
+        throwIO . ErrorCall . unlines $
+          [ "The field csrf_secret is currently " <>
+            if isNothing csrfSecret then "missing." else "invalid."
+          , "I just securely generated such a secret for you:"
+          , "  " <> show freshCsrfSecret
+          , ""
+          , "You can either use the field csrf_secret field from the configuration, or the"
+          , "CSRF_SECRET environement variable, or the command line flag --CSRF_SECRET."
+          , "Beware of keeping this value private, hence you might not want to put it in"
+          , "a configuration file which is version controlled."
+          ]
 
     logger INFO "Press ^C to abort."
     let run = do
--- a/thentos-core/src/Thentos/Frontend/CSRF.hs
+++ b/thentos-core/src/Thentos/Frontend/CSRF.hs
@@ -1,0 +1,99 @@
+{-# LANGUAGE DataKinds             #-}
+{-# LANGUAGE FlexibleContexts      #-}
+{-# LANGUAGE OverloadedStrings     #-}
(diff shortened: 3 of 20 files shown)


fisx:
Thanks, this looks really nice!  If anybody wants to grab it for review please feel free to do so.  I'll try to take a closer look before Monday, too.

thentos-core/src/Thentos/Frontend/CSRF.hs
```diff
@@ -0,0 +1,99 @@
...
+            return $ CsrfToken . cs $ rnd <> convertToBase Base16 (hmac key (tok <> rnd) :: HMAC SHA256)
+          where
+            tok = cs $ fromThentosSessionToken sessionToken
+
+-- | Extracts the nonce part from the CSRF token.
+csrfNonceFromCsrfToken :: CsrfToken -> CsrfNonce
+csrfNonceFromCsrfToken = CsrfNonce . SBS.take 64 . cs . fromCsrfToken
+
+-- | Checks given a 'CsrfToken'.
+-- This token should from the form data of the POST request.
```
Mikolaj: Don't understand.

np:
Thinking about the tests now, given of of the entry points:

* the form should contain a well-form CSRF token
* getting the form twice yield two different tokens
* posting the two forms should work in any order
* removing the CSRF token should make the form rejected
* altering the CSRF token should make the form rejected
* using the CSRF token from another session should make the form rejected

thentos-core/src/Thentos/Frontend/CSRF.hs
np: Indeed...

Verify the authenticity of a given 'CsrfToken'. This token should come from the form data of the POST request NOT the one from the 'FrontendSessionData'.

fhartwig:
Very nice :+1:

Mikolaj:
Pasting the explanation for multiple tokens that are yet valid for the whole session:

20:42 < npouillard> you've been pin-pointing at the right questions and fisx gave you all the answers but one
20:43 < npouillard> so yes tokens are changing at every request but for the server what matter is only per-session
20:44 < npouillard> indeed this means that if the attacker could get access to one of these tokens it would be enough
20:46 < npouillard> these tokens changing at every request is to prevent another attack where one can guess the token by repeatedly querying the server pretending the client. If the server where to send the same token over and over then 
                    the attacker could recover it in some circumstances.
20:47 < npouillard> first all of this is assuming the use of TLS otherwise there is no point bothering about security
20:48 < npouillard> moreover this attack only takes place when HTTP compression is enabled (TLS compression is longer considered secure for this reason)
20:48 < npouillard> the issue is when combining encryption (which hides the contents but not the length) and compression (which makes the length depend on the content)
20:49 < npouillard> if you compress after encryption its safe but useless
20:49 < npouillard> if you compress then encrypt (which is often done), then some data leaks through the length of the ciphertext
20:51 < npouillard> the BEAST attack exploited this by guessing the CSRF token by sending request to the server pretending to be the client injecting in the request the guess of the token in a parameter which is echoed back by the server 
                    to the real client encrypted for the client
20:56 < npouillard> assuming the CSRF token is given as an attribute such as csrf:SOMESECRET to keep it simple, then the guess is going to be csrf:XYZ with all combination of XYZ then you look at which answer was the shortest it is highly 
                    likely that XYZ=SOM will compress better than the rest because of the repetition with the real secret also part of the response. You then proceed with your guess being csrf:SOMXYZ. On a test
20:56 < npouillard> setup they could recover the full token in 30 secs.

Agreed, thank you for the explanation.

thentos-core/src/Thentos/Frontend/CSRF.hs
fisx: are you planning to add this comment to the code instead of the other one?  it's a log clearer!  (-:

fisx:
yes, thanks for the explanation.  would it make sense to have an edited version of this in the code?  i'd be happy to add it.

thentos-core/devel.config
```diff
@@ -1,3 +1,7 @@
+# This "secret" is hardcoded and public.
+# If you remove it, thentos will generate a new one for you.
+csrf_secret: 1daf3741e8a9ae1b39fd7e9cc7bab44ee31b6c3119ab5c3b05ac33cbb543289c
```
fisx: if this can be generated implicitly, then what's the reason it is in the config?  so that form submission works in between server restarts?

fisx: i've just read on a few more lines.  no need to reply.  (-:

nice!

thentos-core/src/Thentos.hs
```diff
@@ -72,6 +73,23 @@ makeMain commandSwitch =
...
+    when (isJust mFeConfig && not (validFormatCsrfSecretField csrfSecret)) $ do
+        CsrfSecret freshCsrfSecret <- genCsrfSecret
+        throwIO . ErrorCall . unlines $
+          [ "The field csrf_secret is currently " <>
+            if isNothing csrfSecret then "missing." else "invalid."
+          , "I just securely generated such a secret for you:"
+          , "  " <> show freshCsrfSecret
+          , ""
+          , "You can either use the field csrf_secret field from the configuration, or the"
+          , "CSRF_SECRET environement variable, or the command line flag --CSRF_SECRET."
```
fisx: did you test the command line flag?  i thought that was `--csrf-secret`.

thentos-core/src/Thentos/Frontend/CSRF.hs
```diff
@@ -0,0 +1,99 @@
+{-# LANGUAGE DataKinds             #-}
+{-# LANGUAGE FlexibleContexts      #-}
+{-# LANGUAGE OverloadedStrings     #-}
+{-# LANGUAGE PackageImports        #-}
+{-# LANGUAGE ScopedTypeVariables   #-}
+{-# LANGUAGE TypeOperators         #-}
+module Thentos.Frontend.CSRF where
```
fisx: i just added an explicit export list that's probably incomplete.  (will push soon!)

thentos-core/src/Thentos/Frontend/State.hs
```diff
@@ -137,21 +127,21 @@ enterFAction aState key smap = Nat $ ExceptT . (>>= _Left fActionServantErr) . r
     run fServer = fst <$> runActionE emptyFrontendSessionData aState fServer'
       where
         fServer' :: FAction a
-        fServer' = do
+        fServer' =
             case smap key of
-                Nothing -> error $ "enterFAction: internal error in servant-session: no cookie!"
+                Nothing ->
+                    -- Should really this be 'error' here.
+                    -- Can a misbehaving client trigger this error? Thus crashing the server?
```
fisx: Isn't `error` thread-local?  Not sure, but no matter: The error message somewhat obscurely tries to state that servant-session is supposed to always come up with a state, whether freshly created or read from the cookie.  I'll update the comment.  (Please check if this sounds plausible.)

fisx: nope, changed my mind: you are right and this should be a more graceful failure.  somehting like `redirect /no-cookie`.

I'll add a FIXME.

np:
Thanks for the reviews and improvements.

np:
Sure @fisx

fisx:
I disabled a breaking test and recorded the coincidental cleanup work that happened during debugging.

I consider the test dispensible for now, since the thentos frontend is prototypical and we're not using it anywhere for now.

thentos-core/src/Thentos.hs
Mikolaj: Reading configifier sources, this may even be thentos_CSRF_SECRET or something similar, and I actually didn't see code switching it into upper case. So a (manual) test would be very welcome.

fisx:
LGTM.

Ready to merge?

Mikolaj:
I'm still not sure about --CSRF_SECRET, see above.

PR #492, opened 2016-02-10
Add submodules/servant/servant-lucid to cabal sandbox.

fisx:
this is helpful if software depending on thentos-core wants
to depend on servant-lucid and our dev snapshot of servant
at the same time.


andorp:
LGTM

PR #498, opened 2016-02-24
Remove unnecessary restriction to Identity monad in Servant.Missing

fisx:
For this to work, we will have to go through the code and replace all occurrances of `Html ()` with `HtmlT m ()`.  Not sure about the implications of that.

### DIFF
--- a/thentos-core/src/Servant/Missing.hs
+++ b/thentos-core/src/Servant/Missing.hs
@@ -17,7 +17,6 @@ module Servant.Missing
   ,formRedirectH) where
 
 import Control.Monad.Except (MonadError, throwError)
-import Control.Monad.Identity (Identity, runIdentity)
 import Data.Bifunctor (first)
 import Data.Proxy (Proxy(Proxy))
 import Data.String.Conversions (ST, cs)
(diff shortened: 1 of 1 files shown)


fisx:
(I guess I'm just opening this to get opinions and make sure the idea isn't lost.  I stumbled into this when trying to create html from generic forms over at liqd/aula.)

fisx:
I fixed the problem I had in https://github.com/liqd/aula/pull/104, so this PR gets downgraded from "would be really nice to have" to "curiosity".

fisx:
much simpler, thanks!  (-:

from irc, just for the record:

```
11:20 < npouillard> fisx_: about the thentos PR, your change does not really relax (but add) a restriction, however with a slight change we can get what you wanted
11:20 < npouillard> in a way `Env Identity` is quite generous as it is isomorphic to `forall f. Applicative f => Env f`
11:21 < npouillard> however GHC chokes at this kind of type family so this would need a wrapper type...
11:21 < npouillard> instead we just provide a function: fromEnvIdentity :: Applicative m => Env Identity -> Env m
11:22 < npouillard> which can be used when the Env you get is not of the right functor/monad
11:22 < fisx_> so this is the function i implemented by hand in exec/RenderHtml.hs now.
```

PR #501, opened 2016-02-29
Simple wai file upload for digestive-functors.

fisx:


thentos-core/src/Servant/Missing.hs
```diff
@@ -41,6 +42,12 @@ data FormReqBody
 fromEnvIdentity :: Applicative m => Env Identity -> Env m
 fromEnvIdentity env = pure . runIdentity . env
 
+-- | FIXME: not sure this is legal.  do we need to make the IORef thread safe?  by wrapping it into
+-- an STRef?  but it seems like we are using it exactly like it is intended to.
```
Mikolaj: The single, unprotected IORef seems correct --- looking at implementation of resourceForkIO, we see only a single IORef is passed to all the resourceForkIO-spawned threads and since it's a single IORef, the atomicModifyIORef wrappers, that it's protected with, are enough to prevent race conditions, etc. The question remains whether we should use/reimplement resourceForkIO. Investigating...

Mikolaj: Also, the InternalState should be closed somewhere, using closeInternalState, or else we need to call runResourceT from time to time. Otherwise, no releasing of resources ever happens, because tempFileBackEnd doesn't do that for us. Oh, you actually mention that in the FIXME below.

fisx: > Also, the InternalState should be closed somewhere, using closeInternalState, or else we need to call runResourceT from time to time. Otherwise, no releasing of resources ever happens, because tempFileBackEnd doesn't do that for us. Oh, you actually mention that in the FIXME below.

resourcet is interesting!  can you offer an implementation of the functionality here that cleans up after itself?

fisx:
What happens in the common case that we don't upload any files if we don't call the clear state action?

Would it be a good idea to, before returning the `FormData`, check if the file list is empty, and if so, call the cleanup action right away and return nothing but `pure ()`?

fisx:
LGTM!

I am still interested in discussion the above question, but I'll merge this first.  Thanks!

np:
@fisx are you concerned by some memory leak of holding too long on this internal state?

fisx:
> @fisx are you concerned by some memory leak of holding too long on this internal state?

yes, sort of.  but after off-line discussion, i say we wait and see.

Mikolaj:
@np, you nailed it (and @fixs managed to even improve on it). 

Re memory leak, I'd not worry, the internal state is tiny, doesn't refer to anything in our code in its thunk (even if we call `resourceForkIO`) and, in the absence of actual files, it would grow only if started calling `resourceForkIO` and even then only a little.

PR #504, opened 2016-03-16
Servant.Missing: prismatic error handling

np:


fisx:
Are these changes compatible with aula, or will we need to bump the thentos-core dependency there?

fisx:
[@np on irc]

intutively `#` is the `$` of prisms, so `_Just # x` does the same `Just x`, of course the prism at hand could some smart constructor for which you don't have/want the normal version.  also the docs says `#` is an infix alias for `review`; `review` being equivalent to `view . re`

finally 're' is what can turn around a prism. If we think of

```haskell
_Just :: Prism' (Maybe a) a
```

it goes from the outer type `Maybe a` to the inner type `a`. From left to right, a prism is a matcher (a case). From right to left a prism is a constructor.

```haskell
re _Just :: Getter a (Maybe a)
```

fisx:
LGTM.

I couldn't find any compatibility issues with aula locally, so I will merge this.  Thanks!

PR #506, opened 2016-04-19
Allow lists of content-types in FormH, not just one.

fisx:
We need this in liqd/aula to deliver PlainText versions of
form pages for debugging and development.


np:
LGTM

PR #507, opened 2016-06-30
Upgrade to lts-6.3

np:


fisx:
LGTM, thanks!  I moved configifier from submodule to hackage and bumped the version.  Please merge if you agree and travis is green.

fisx:
oops.

fisx:
(the hlint error is a bit weird, but i think working around it is better than keeping the hlint dependency hack that i removed in 453f229.)

